PR #54, opened 2020-02-17
bs_theme_quick feature

jcheng5:
## Goal

Allow users to specify a _very_ small number of colors and have the entire theme change to reflect that. Currently the only required colors are `bg` and `fg`, with optional `accent` (primary) and `secondary` colors.

## Example

```r
bs_theme_quick(bg = "black", fg = "white", accent = "red")
```

image

## Limitations

- **`dark` and `light` lose their intuitive meanings.** If you specify a dark bg and light fg, then `dark` and `light` mean "closer to bg" and "closer to fg", not luminosity.

## TODO

- [x] Implement bs3 version
- [x] Unit tests
- [x] Remove dependency on `shiny:::parseCssColor` (move to htmltools?)

## Open questions

- Should the function actually add the theme, or just generate it? **The exported bs_theme_quick function actually adds it; the underlying (un-exported) functions bs4_theme_quick and bs3_theme_quick just generate.**
- Add fonts? (Seems like the next most common thing that will come up) We could do just one font, or separate fonts for body, heading, buttons/inputs...  **Not as part of this function**


R/quick-theme.R
```diff
@@ -0,0 +1,210 @@
...
+    ))
+  }
+  if (white_yiq < black_yiq) {
+    # Invert yiq colors
+    results <- c(results, list(
+      "yiq-text-light" = grays[["gray-900"]],
+      "yiq-text-dark" = grays[["white"]]
+    ))
+  }
+
```
cpsievert: Add `!default` flags here

R/quick-theme.R
```diff
@@ -0,0 +1,210 @@
...
+    `carousel-indicator-active-bg` = "$white", `carousel-indicator-border-color` = "$white",
+    `carousel-caption-color` = "$white", `close-text-shadow` = "$white",
+    `kbd-color` = "$white")
+
+  results <- c(result_colors, color_mapping)
+
+  if (!is.null(accent)) {
+    results <- c(results, list("brand-primary" = accent))
+  }
+
```
cpsievert: Add `!default` flags

R/quick-theme.R
jcheng5: I don't think it's needed? This list is going into `sass_layer(defaults = ...)`

cpsievert: `sass_layer(defaults = ...)` won't automatically add a `!default` flag...only `bs_theme_add_variables()` will at the moment

R/quick-theme.R
```diff
@@ -0,0 +1,210 @@
...
+bs_theme_quick <- function(bg = "#FFFFFF", fg = "#000000",
+  accent = NULL, secondary = NULL) {
+
+  theme <- bs_theme_get()
+  if (is.null(theme)) {
+    stop("No bootstraplib theme is active (did you forget to call bs_theme_new()?)")
+  }
+
+  bs4_tags <- c("bootstraplib_version_4", "bootstraplib_version_4+3")
+  bs3_tags <- c("bootstraplib_version_3")
```
cpsievert: Any particular reason why you're not using `theme_version()` here?

R/quick-theme.R
```diff
@@ -0,0 +1,210 @@
...
+
+  if (any(grays[,4] != 255)) {
+    warning(call. = FALSE,
+      "bs_theme_quick does not respect alpha in `white` and `black` arguments"
+    )
+  }
+
+  white_yiq <- color_yiq(grays[1,1], grays[1,2], grays[1,3])
+  black_yiq <- color_yiq(grays[11,1], grays[11,2], grays[11,3])
+
```
cpsievert: What do you think about warning if the contrast ratio is less than 7 and pointing to this website (I was thinking of doing something like this for the `autoTheme` feature as well)? https://webaim.org/resources/contrastchecker/

R/quick-theme.R
jcheng5: Oooh good catch then. I fixed and added a test.

R/quick-theme.R
jcheng5: I forgot about it!

R/quick-theme.R
jcheng5: Great idea! I don't think I'll have time to implement it though--at least not today, I only have a few more minutes of coding time!

What if we let this PR go through without it, and if/when you add it to autoTheme, then we add it here as well?

cpsievert: Sounds good, I'd still like some more time to review, but if no major blockers I'll review and merge today

PR #57, opened 2020-02-22
bs_theme_accent_colors() function

jcheng5:
Besides adding the `bs_theme_accent_colors()`, I extracted out some common pieces between the existing base color setter and the new accent colors setter:

1. Validate that `bs_theme_new` was called
2. Dispatch to the correct implementation based on `theme_version()` (and error if we don't know about this version)
3. Drop/ignore NULLs
4. Validate colors
5. Add `!default`
6. Create sass layer and `bs_theme_add`

This is all handled in `dispatch_theme_setter` now (except color validation, you have to call that separately, since this needs to work for fonts too).

## TODO

- [x] Reference documentation for `bs_theme_accent_colors()`
- [x] Update documentation for `bs_theme_base_colors()` to remove `accent` and `secondary`
- [x] Unit tests for lower-level functions `format_varnames`, `validate_and_normalize_colors`, `dispatch_theme_setter`
- [x] Unit tests that cover all accent colors


R/theme-base-colors.R
```diff
@@ -203,11 +174,202 @@ bs3_theme_base_colors <- function(bg, fg, accent, secondary) {
...
+  # Warns and filters out unsupported arguments
+  supported <- c("primary", "success", "info", "warning", "danger")
+  matches <- match(names(args), supported)
+  bad_names <- names(args)[is.na(matches)]
+  if (length(bad_names) > 0) {
+    warning(call. = FALSE, "Bootstrap 3 doesn't support the following accent ",
+      "color argument(s), they will be ignored: ", format_varnames(bad_names)
+    )
+  }
+  args <- args[!is.na(matches)]
```
cpsievert: Might be useful to abstract this logic out into it's function?

R/theme-base-colors.R
```diff
@@ -35,45 +26,43 @@
...
 
-bs4_theme_base_colors <- function(bg, fg, accent, secondary) {
-  white <- htmltools:::parseCssColors(bg)
-  black <- htmltools:::parseCssColors(fg)
+bs4_theme_base_colors <- function(args) {
+  white <- args$bg
+  black <- args$fg
 
-  grays <- colorRamp(c(white, black), alpha = TRUE)(0:10/10)
+  grays <- grDevices::colorRamp(c(white, black), alpha = TRUE)(0:10/10)
```
cpsievert: This could be for another PR if you want, but I think we should be using `scales::colour_ramp()`

cpsievert:
LGTM (other than the comments) 👍

cpsievert:
Btw, do you think it'll be worthwhile to have some thing like `bs_theme_fonts()` for `font-size`? If so, maybe `bs_theme_fonts()` should be `bs_theme_font_family()`?

R/theme-base-colors.R
jcheng5: I'm going to hold off on this right now, it's not a drop-in replacement because it returns `#RRGGBB` results instead of a color matrix; we'd have to at least do `col2rgb` on the result.

jcheng5:
@cpsievert I went to `bs_theme_font_family` but didn't like the look of it, switched back to `bs_theme_font`. I didn't think this would be unreasonable if we hypothetically had a `bs_theme_font_size` someday:

```r
bs_theme_font(heading = "Impact")
bs_theme_font_size(heading = "18pt")
```

or something more like

```r
bs_theme_font(heading = font_options("Impact", "18pt"))
```

R/theme-base-colors.R
```diff
@@ -203,11 +174,312 @@ bs3_theme_base_colors <- function(bg, fg, accent, secondary) {
...
+    str
+  )
+
+  paste(str, collapse = ", ")
+}
+
+#' Remove unsupported arguments, with a nicely formatted warning
+#'
+#' @param caller_name String naming the calling function; used for error
+#'   messages
```
cpsievert:
```suggestion
#'   messages
#' @param arg_name A name for the group of variables (e.g., accent, font); used for error messages
```

R/theme-base-colors.R
```diff
@@ -203,11 +174,312 @@ bs3_theme_base_colors <- function(bg, fg, accent, secondary) {
...
+#' ```
+#'
+#' The resulting CSS will contain `font-family: Source Sans Pro, sans-serif;`
+#' which is technically out of spec, but in fact is likely to still work with
+#' most browsers.
+#'
+#' @param base The default typeface.
+#' @param code The typeface to be used for code. Be sure this is monospace!
+#' @param heading The typeface to be used for heading elements.
+#'
```
cpsievert:
```suggestion
#' @family customizations
```

PR #67, opened 2020-03-17
Have themer try it's best to disable the initial bootstrap stylesheet

cpsievert:
## reprex of issue

```r
library(bootstraplib)
bs_theme_preview()
```

Disable the "rounded corners" option. Notice how the well and pills still have rounded corners. The reason is that with `$enable-rounded: false`, `border-radius` isn't defined; however, since the initial Bootstrap CSS is built with `$enable-rounded: true`, the rules from the initial build still apply

### DIFF
--- a/R/theme-preview.R
+++ b/R/theme-preview.R
@@ -111,7 +111,9 @@ bs_themer_ui <- function() {
   withTags(tagList(
     colorpicker_deps(),
     htmlDependency(
-      "bs_themer", version = packageVersion("bootstraplib"), src = "themer", script = "themer.js", package = "bootstraplib", all_files = FALSE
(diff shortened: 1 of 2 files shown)


jcheng5:
Is this a pretty cross platform approach?

cpsievert:
Oh, good catch, it doesn't work for IE11, but neither does the colorpicker at the moment...

cpsievert:
Maybe a dynamic UI approach would be better? I did start going down that road initially, but then stopped when I realized we'd probably need users to place both `bs_themer_ui()` in the ui and `bs_themer()` in the server

jcheng5:
If it's only IE11, I'm fine with that for this particular feature. I was more concerned it didn't work for the more modern ones.

PR #70, opened 2020-04-23
Updates for thematic

cpsievert:
To be merged after thematic is made public.

Only real code change is 522a870


DESCRIPTION
```diff
@@ -34,7 +34,7 @@ Suggests:
 License: MIT | file LICENSE
```
jcheng5, line 34: Suggest `thematic`?

PR #88, opened 2020-08-27
Use consistent output directory for bootstrap()

wch:
This PR does the following:

* When the inputs are the same, it uses the same output directory every time. If the output directory already exists, it won't bother writing files there.
* Adds a wrapper function `system_file()`, which is faster than `system.file`. (The latter was a bottleneck.)

Before the `system_file` change:
```
> system.time(bootstrap())
   user  system elapsed 
  0.403   0.038   0.441 
> system.time(bootstrap())
   user  system elapsed 
  0.034   0.009   0.043
```

After:

```
> system.time(bootstrap())
   user  system elapsed 
  0.385   0.032   0.416 
> system.time(bootstrap())
   user  system elapsed 
  0.010   0.003   0.013 
```

This should be merged after https://github.com/rstudio/sass/pull/44. (The `Remotes` field here will have to be updated after that PR is merged.)

*****

Testing notes:

Run the following. Each time you run `dir(outdir)`, the output should be the same -- there should be exactly one directory listed with a name like `"bootstrap-xxxxxxx"`. If there is more than one such directory, then there's a problem.

```R
library(bootstraplib)
outdir <- file.path(tempdir(), "bootstraplib")

bootstrap()
dir(outdir)
#> [1] "bootstrap-560e344a81bc5cbe"

bootstrap()
dir(outdir)
#> [1] "bootstrap-560e344a81bc5cbe"
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -53,4 +53,4 @@ Collate:
 Remotes:
     rstudio/thematic,
     rstudio/jquerylib,
-    rstudio/sass@disk-cache
+    rstudio/sass
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -46,9 +46,9 @@ Collate:
     'layers.R'
     'theme-add.R'
     'theme-custom.R'
+    'utils.R'
     'theme-preview.R'
     'themes.R'
-    'utils.R'
     'versions.R'
 Remotes: 
     rstudio/thematic,
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -22,7 +22,7 @@ Imports:
     grDevices,
     htmltools (>= 0.4.0.9003),
     jsonlite,
-    sass (>= 0.2.0),
+    sass (>= 0.2.0.9001),
     jquerylib (>= 0.1.2),
     rlang
 Suggests:
@@ -35,7 +35,7 @@ Suggests:
 License: MIT | file LICENSE
 Encoding: UTF-8
 LazyData: true
-RoxygenNote: 7.1.0
+RoxygenNote: 7.1.1
 Roxygen: list(markdown = TRUE)
 VignetteBuilder: knitr
 Collate: 
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -52,5 +52,5 @@ Collate:
     'versions.R'
 Remotes:
     rstudio/thematic,
-    rstudio/jquerylib
+    rstudio/jquerylib,
     rstudio/sass@disk-cache
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -18,7 +18,7 @@ Authors@R: c(
     )
 Description: Tools for obtaining bootstrap CSS and customizing it via SASS.
 Depends: R (>= 2.10)
-Imports: 
+Imports:
     grDevices,
     htmltools (>= 0.4.0.9003),
     jsonlite,
@@ -38,7 +38,7 @@ LazyData: true
 RoxygenNote: 7.1.1
 Roxygen: list(markdown = TRUE)
 VignetteBuilder: knitr
-Collate: 
+Collate:
     'bootswatch.R'
     'bs_sass.R'
     'files.R'
@@ -50,6 +50,7 @@ Collate:
     'theme-preview.R'
     'themes.R'
     'versions.R'
-Remotes: 
+Remotes:
     rstudio/thematic,
     rstudio/jquerylib
+    rstudio/sass@disk-cache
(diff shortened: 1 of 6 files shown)


DESCRIPTION
```diff
@@ -35,21 +35,22 @@ Suggests:
...
     'theme-custom.R'
+    'utils.R'
     'theme-preview.R'
     'themes.R'
-    'utils.R'
     'versions.R'
-Remotes: 
+Remotes:
     rstudio/thematic,
     rstudio/jquerylib
```
cpsievert, line 55:
```suggestion
    rstudio/jquerylib,
```

R/bs_sass.R
```diff
@@ -61,23 +61,34 @@ bootstrap <- function(theme = bs_theme_get(),
   )
   opts <- utils::modifyList(opts, options)
 
-  # Temp dir for building the HTML dependencies
-  output_path <- tempfile("bscustom")
-  dir.create(output_path)
+  cache_key <- sass::sass_cache_key(theme, opts)
```
cpsievert, line 64: We'll should incorporate the bootstrap version somehow here (which is determined below by `if (version %in% "3") version_bs3 else version_bs4`)

R/bs_sass.R
```diff
@@ -61,23 +61,34 @@ bootstrap <- function(theme = bs_theme_get(),
...
-    input = theme,
-    output = file.path(output_path, output_css),
-    write_attachments = TRUE,
-    options = opts,
-    ...
+  # Temp dir for building the HTML dependencies
+  output_path <- file.path(
+    tempdir(),
+    "bootstraplib",
+    paste0("bootstrap-", cache_key)
```
cpsievert, line 70: @schloerke I vaguely remember you running into learnr issues with filenames that were too long -- could that be an issue here?

wch: IDK if it matters, but the hash isn't too long. For example, the dir will look like `bootstrap-52a853bb558aa1b2`.

R/bs_sass.R
wch: Sounds good -- I've extracted this out into a function.

R/theme-preview.R
```diff
@@ -43,12 +46,16 @@ colorpicker_deps <- function() {
   )
 }
 
-opts_metadata <- jsonlite::fromJSON(system.file("themer/options.json", package = "bootstraplib"),
```
wch, line 46: Using `system_file` saved us from a bad build-time bug that would have bitten us after it went to CRAN. 😬 

When I was building and installing the package after adding `system_file`, I started getting this error:

```
** testing if installed package keeps a record of temporary installation path
ERROR: hard-coded installation path: please report to the package maintainer and use ‘--no-staged-install’
```

It's because the package is calling `system_file()` at build time, and the local path is getting stored in `system_file`'s cache.

The problem actually did occur before I added `system_file`, but because the local path got stored as part of a JSON object, R didn't detect that the path got embedded in there. With the path cached in `system_file`, it was able to detect the problem. Whew!

R/bs_sass.R
schloerke: That should be fine. For learnr it was like a few hundred characters on windows for some reason.  But that was because when normalized on a university machine, it resolved to a path that was really convoluted.  I also think for the learnr situation it was caused by paste'ing two paths together.  For now, I wouldn't worry about it. 

If anything you could drop the `bootstrap-` prefix on the file as it is already namespaced by the `bootstrap` folder.

R/theme-preview.R
schloerke: Good catch!!

R/utils.R
```diff
@@ -63,3 +67,21 @@ color_yiq <- function(r, g, b) {
...
+  package_dir_cache <- character()
+
+  function(..., package = "base") {
+    if (package %in% names(package_dir_cache)) {
+      package_dir <- package_dir_cache[[package]]
+    } else {
+      package_dir <- system.file(package = package)
+      package_dir_cache[[package]] <<- package_dir
+    }
+
```
cpsievert, line 84: This could be problematic if we ever want to use other named args `system.file()`...perhaps we can just add `ellipsis::check_dots_unnamed()`?

DESCRIPTION
```diff
@@ -35,21 +35,22 @@ Suggests:
...
     'theme-preview.R'
     'themes.R'
-    'utils.R'
     'versions.R'
-Remotes: 
+Remotes:
     rstudio/thematic,
-    rstudio/jquerylib
+    rstudio/jquerylib,
+    rstudio/sass@disk-cache
```
cpsievert, line 56: Once https://github.com/rstudio/sass/pull/44 is merged

```suggestion
    rstudio/sass
```

R/utils.R
wch: Good thinking. I'll implement something similar so we don't have to bring in the ellipsis package.

cpsievert (approved):
LGTM (once conflicts are resolved)

PR #90, opened 2020-09-04
Leverage the new sass::output_file()

cpsievert:
Requires https://github.com/rstudio/sass/pull/48

### DIFF
--- a/R/bs_sass.R
+++ b/R/bs_sass.R
@@ -61,7 +61,7 @@ bootstrap <- function(theme = bs_theme_get(),
   out_file <- sass::sass(
     input = theme,
     options = sass_options,
-    output = sass::output_file(basename = "bootstrap", dirname = "bootstraplib-"),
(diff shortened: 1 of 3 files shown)


R/bs_sass.R
```diff
@@ -57,53 +57,38 @@ bootstrap <- function(theme = bs_theme_get(),
...
-    dir.create(output_path, recursive = TRUE)
-
-    # Compile sass in temp dir
-    sass::sass(
-      input = theme,
-      options = sass_options,
-      output = file.path(output_path, output_css),
-      cache = cache,
-      write_attachments = TRUE
+  out_file <- sass::sass(
```
wch, line 61: The logic has changed here so that it always overwrites the files in the bootstraplib-xxx directory, right? It would be good to check if there's a noticeable performance penalty.

wch (approved):

wch (approved):
Looks good. Looks like some CI failure due to docs?

PR #104, opened 2020-09-16
Have bs_theme_new() call shinyOptions(bootstraplib = T)

### LINKED ISSUE #103: Provide a way to use theme shiny without effecting global state

cpsievert:
Closes #103

### DIFF
--- a/R/shiny.R
+++ b/R/shiny.R
@@ -1,0 +1,25 @@
+.globals <- new.env(parent = emptyenv())
+
+shiny_options_set <- function() {
+  do_shiny({
+    .globals$shinyOptions <- shinyOptions()
+    shiny::shinyOptions(bootstraplib = TRUE)
+  })
+}
+
+shiny_options_restore <- function() {
+  do_shiny({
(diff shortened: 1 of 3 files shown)


R/shiny.R
```diff
@@ -0,0 +1,25 @@
...
+    shiny::shinyOptions(bootstraplib = .globals$shinyOptions$bootstraplib)
+  })
+}
+
+do_shiny <- function(expr) {
+  if ("shiny" %in% loadedNamespaces()) {
+    return(expr)
+  }
+
+  setHook(
```
wch, line 21: I think that calling `setHook` this way has a couple issues: it could create an indefinitely large queue of callbacks to call when shiny is loaded. If the user has no intention of even using Shiny in this session, that is an unnecessary use of memory.

The other issue is that if/when shiny actually is loaded, this will execute many of these stored-up callbacks, which isn't necessary. We really only care about making the most recent bootstrap theme used by shiny.

Probably better do something like this: in `.onLoad`, check if shiny is loaded; if not, call `setHook()`. What that hook will do (if shiny is loaded later) is call call `shinyOptions()`.

PR #106, opened 2020-09-16
'Intelligent' management of global state

cpsievert:
(To be merged after https://github.com/rstudio/shiny/pull/3062)

This PR makes it so that running a shiny app via file (i.e., `runApp()`) won't alter bootstraplib's global state. That is, the following should always be `TRUE`:

```r
theme <- bs_theme_get()
shiny::runApp()
identical(theme, bs_theme_get())
#> TRUE
```

On the other hand, if the app's source is copy/pasted into the R console, then the bootstraplib theme will always be cleared after the app exits:

```r
bs_theme_new()
shinyApp(fluidPage(), function(input, output) {})
bs_theme_get()
#> NULL
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -39,6 +39,7 @@ RoxygenNote: 7.1.1
 Roxygen: list(markdown = TRUE)
 VignetteBuilder: knitr
 Collate:
+    'globals.R'
     'bootswatch.R'
     'bs_sass.R'
     'files.R'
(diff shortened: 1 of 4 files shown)


cpsievert:
Ack, I just realized that, in the case that `app.R` _doesn't_ call `bs_theme_new()`, this is the current behavior:

```r
bs_theme_new()
shiny::runApp("tmp")
bs_theme_get()
#> NULL
```

which clearly isn't right...

cpsievert:
After a bit more thought, I think it'll make more sense to manage bootstraplib's state inside of shiny

PR #107, opened 2020-09-17
Use sass_default_cache()

wch:
This should be merged after https://github.com/rstudio/sass/pull/50. It uses the updated API introduced by that PR.


cpsievert:
@wch this one can be closed, right?

PR #109, opened 2020-09-17
Include pre-compiled copies of stock Bootstrap CSS

wch:
(To be merged after https://github.com/rstudio/sass/pull/50)

This adds precompiled builds of stock Bootstrap, for `bs_theme_create("4")`, `bs_theme_create("4+3")`, and `bs_theme_create("3")`.

* If the theme is modified at all (as with `bs_theme_add()`), or `bootstrap()` is called with any non-default `sass_options`, then it will not use the precompiled versions.
* The precompiled CSS files are generated by tools/build-css-prebuilt.R
* The tests will fail if the pre-built build differs from the compiled build, including attachments. This will ensure that, if we update any sass files, we will be forced to update the pre-built CSS files.


tools/README.md
```diff
@@ -7,6 +7,7 @@ Rscript tools/yarn_install.R
 ```
 
 **Warning:** Upgrading to a new major version of Bootstrap won't be as simple as running this R script. We'll cross that bridge when we get there.
+Note that after upgrading Bootstrap, you should generate the precompiled CSS files by running the precompile-css.R script in this directory. (Eventually this step should be added to the Bootstrap upgrade script.)
```
cpsievert, line 10: Is there any reason why we can't do this at the end of the `tools/yarn_install.R` script (instead of having a separate build script)? I'm worried we'll forget to run this and wind up shipping stale versions of Bootstrap

tests/testthat/test-precompile.R
```diff
@@ -0,0 +1,83 @@
+# Disable caching
+sass_cache_set(NULL)
```
cpsievert, line 2: Can we do this locally (for this test script)?

tests/testthat/test-precompile.R
```diff
@@ -0,0 +1,83 @@
...
+
+  dir1 <- get_bootstrap_path(bootstrap(bs_theme_create("4+3"), use_precompiled_css = FALSE))
+  dir2 <- get_bootstrap_path(bootstrap(bs_theme_create("4+3"), use_precompiled_css = TRUE))
+  expect_false(identical(dir1, dir2))
+  expect_true(identical_dirs(dir1, dir2))
+
+  dir1 <- get_bootstrap_path(bootstrap(bs_theme_create("3"), use_precompiled_css = FALSE))
+  dir2 <- get_bootstrap_path(bootstrap(bs_theme_create("3"), use_precompiled_css = TRUE))
+  expect_false(identical(dir1, dir2))
+  expect_true(identical_dirs(dir1, dir2))
```
cpsievert, line 44:
```suggestion
  for (version in c("3", "4", "4+3")) {
    dir1 <- get_bootstrap_path(bootstrap(bs_theme_create(version), use_precompiled_css = FALSE))
    dir2 <- get_bootstrap_path(bootstrap(bs_theme_create(version), use_precompiled_css = TRUE))
    expect_false(identical(dir1, dir2))
    expect_true(identical_dirs(dir1, dir2))
  }
  
```

tools/README.md
wch: I didn't add it there because the comment said "Upgrading to a new major version of Bootstrap won't be as simple as running this R script", so I assumed that it wasn't safe to put there. But if you think it can go there, I'm fine with that. Note that the package needs to be loaded to run the script.

FWIW, the unit tests ensure that the precompiled css matches the normally compiled css. I tested this out by modifying one of the scss files and running the unit tests. The tests failed, which is good.

tests/testthat/test-precompile.R
wch: Yeah, I actually have this here just until we finalize and merge the sass caching changes.

R/themes.R
```diff
@@ -36,6 +36,9 @@
 #' # Helpful if you want to know the Bootstrap version of the current theme
 #' theme_version()
 #'
+#' # Can create a theme and return the object, instead of setting it globally.
+#' bs_theme_create()
+#'
```
cpsievert, line 41: If we're going to include `bs_theme_create()` in the examples, let's show how one might use it...In fact, let's move this down and replace lines 56-66 with something like:

```r
bs3 <- bs_theme_create(3)
red_default <- sass::sass_layer(defaults = list("brand-primary" = "red !default"))
bs3 <- sass::sass_layer_merge(bs3, red_default)
bootstrap_sass(".my-class { color: $brand-primary; }", bs3)
```

tests/testthat/test-precompile.R
wch: I just remembered: we also have stuff that disables caching in other test files, and those will have to be fixed as well.
https://github.com/rstudio/bootstraplib/blob/4788d871a9bb09a8f700e7cb4d886b161184054f/tests/testthat/test-bs4-deps.R#L4

R/bs_sass.R
```diff
@@ -1,32 +1,40 @@
...
+#'   [sass::sass_layer()] which contains a bootstraplib theme. Useful for adding
+#'   custom layers to the current theme without affecting the global state
+#'   (e.g., `sass::sass_layer_merge(bs_theme_get(), my_layer())`).
 #' @param sass_options see [sass::sass_options()].
 #' @param jquery See [jquerylib::jquery_core()].
+#' @param use_precompiled_css Before compiling the theme object, first look for
+#'   a precompiled CSS file for the given `theme` and `sass_options`. The
+#'   bootstraplib comes with precompiled CSS files for stock Bootstrap builds
+#'   for when `bs_theme_new()` or `bs_theme_create()` are called with `version`
+#'   of `"3"`, `"4+3"`, and `"4"`, and no `bootswatch` theme. If this option is
```
cpsievert, line 29: I'd like to avoid listing supported versions in multiple place (to make it easier to add new versions)

tests/testthat/test-precompile.R
wch: I generally try to avoid having test expectations in functions and loops, because when they do fail, it can be difficult to figure out which condition failed.

For example, the output of testthat looks like this if one of these tests fails:

```
test-precompile.R:39: failure: Precompiled theme output is identical to compiled themes
identical_dirs(dir1, dir2) isn't true.
```

If it's in a loop, we don't know which condition had a problem.

R/versions.R
```diff
@@ -1,5 +1,5 @@
-#' Bootstrap major version default
-#' @export
-version_default <- function() "4+3"
-
-# TODO: make this a getter/setter?
+# DO NOT EDIT
+# This file is auto-generated by tools/yarn_install.R
+version_bs4 <- "4.5.2"
+version_bs3 <- "3.4.1"
```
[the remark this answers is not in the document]
wch: That's a good idea -- want to file an issue?

R/themes.R
wch: Would it be better to have the `bs_theme_add` function take a `.theme` argument, and then make an example with that? (That would be a separate PR)

tools/README.md
cpsievert: Ok, great that we have the test...I think I'd still prefer that we move the logic to `tools/yarn_install.R`...We could even loop over the `versions` as I mentioned in https://github.com/rstudio/bootstraplib/pull/109/files#r491023391

R/bs_sass.R
wch: I don't think the set of precompiled versions is listed elsewhere, is it?

R/themes.R
cpsievert: Maybe, but I still think that it's useful to point out that themes are just sass layers.

BTW, I looked a little into adding a `.theme` argument yesterday and it's not as trivial as I was hoping

R/bs_sass.R
cpsievert: I just mean generally...when we go to add Bootstrap 5 I'd like that to be as seamless as possible. How about we say something like

```r
#' @param use_precompiled_css Before compiling the theme object, first look for
#'   a precompiled CSS file for the given `version`. At the moment, we only provide
#'   precompiled CSS for "stock" builds of Bootstrap (i.e., no theming additions or
#'   non-default `sass_options`).
```

tools/README.md
wch: It's not clear to me exactly what yarn_install.R does with respect to Bootstrap. When we want to update Bootstrap, do we need to manually update Bootstrap first, and then run that script? If so, then I agree that the precompilation stuff can go there.

But if we run the script and then upgrade Bootstrap, then it doesn't make sense to precompile in that script.

R/versions.R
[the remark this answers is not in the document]
cpsievert: Done https://github.com/rstudio/bootstraplib/issues/110

R/precompiled.R
```diff
@@ -0,0 +1,54 @@
...
+#' @return The path to the precompiled CSS file, if it exists, or `NULL`, if
+#'   not.
+#'
+#' @examples
+#' # The path to a precompiled CSS file for a theme
+#' precompiled_css_path(bs_theme_create("4"))
+#'
+#' # Attempting to get the path to a theme that is not precompiled will return NULL
+#' precompiled_css_path(bs_theme_create(bootswatch = "sketchy"))
+#' @export
```
cpsievert, line 46: I feel uneasy about exporting this...if we give them this path, they can easily shoot themselves in the foot by not considering the `file_attachments`

wch: I wanted to have some way that people could check if there was a precompiled versin of their theme. Without this, there's no way for users to check. Any ideas on another way to do it?

cpsievert: I may be missing something, but why should users care? If they really do, they can sniff the `src` path of the returned `htmlDependency()`

tools/README.md
cpsievert: > It's not clear to me exactly what yarn_install.R does with respect to Bootstrap.

It runs `yarn install` with https://github.com/rstudio/bootstraplib/blob/master/inst/package.json, then a bunch of post-processing on the generated files

> do we need to manually update Bootstrap first, and then run that script?

Yea, you currently need to modify the package.json to update or remove the lock file

R/precompiled.R
wch: How about making it `@keywords internal`?

cpsievert: I'd prefer un-exported until there's a request/need for it, but yes, it should at least be `@keywords internal`

R/themes.R
```diff
@@ -93,16 +88,19 @@ bs_theme_set <- function(theme) {
   invisible(old_theme[["bootstraplib_theme"]])
 }
 
-
-bs_theme_create <- function(version = version_default()) {
+#' @rdname theming
+#' @export
+bs_theme_create <- function(version = version_default(), bootswatch = NULL) {
```
cpsievert, line 93: Now that I'm working on a new (functional) interface, I think I'm going to change this to `bs_theme()` (and add more arguments), so I'd prefer that we don't export until that happens

R/precompiled.R
cpsievert: Also, if we do export, we should add a disclaimer to the `@description`

PR #114, opened 2020-09-21
Functional interface

### LINKED ISSUE #111: Make basic theme settings available in bs_theme_new()/bs_theme_create()
Two benefits, you don't have to remember/discover the appropriate function name (e.g., `bs_theme_base_colors()`) and also tab completion

cpsievert:
* Deprecates `bs_theme_*()` functions in favor of `bs_global_*()`
* Introduces a new, functional (& simplified), interface (`bs_theme()`)

Closes #111 and #103

## TODO

- [x] Unit tests
- [x] Update readme
- [ ] Update recipes vignette


R/bs-theme-preview.R
```diff
@@ -459,28 +435,35 @@ bs_theme_get_variables <- function(varnames) {
...
-    print(rlang::expr(`!!`(func_name)(!!!call_vals)))
-  }
 
-  return_vals
+echo_and_eval_theme_code <- function(theme, vals) {
+  funcs <- list(
+    bs_base_colors = bs_base_colors,
+    bs_accent_colors = bs_accent_colors,
+    bs_fonts = bs_fonts
+  )
```
cpsievert, line 445: This could probably be simplified to use the new `bs_theme_update()`. If we can, we should consider not exporting `bs_base_colors`, `bs_accent_colors`, and `bs_fonts`

R/bs-theme.R
```diff
@@ -1,3 +1,158 @@
...
+
+#' @rdname bs_theme
+#' @export
+bs_theme_update <- function(theme = bs_theme(), ..., bg = NULL, fg = NULL,
+                            primary = NULL,  secondary = NULL, success = NULL,
+                            info = NULL, warning = NULL, danger = NULL,
+                            base_font = NULL, code_font = NULL, heading_font = NULL) {
+  theme <- assert_bs_theme(theme)
+  # See R/bs-theme-update.R for the implementation of these
+  theme <- bs_base_colors(theme, bg = bg, fg = fg)
```
cpsievert, line 131: What should happen in the case of `bs_theme_update(bs_theme(bg = "black", fg = "white"), fg = "red")` (should we try to remember the previous `bg`)?

Make sure we also have ample tests of this logic

R/bs-theme.R
```diff
@@ -1,3 +1,158 @@
...
+#'
+#' The resulting CSS will contain `font-family: Source Sans Pro, sans-serif;`
+#' which is technically out of spec, but in fact is likely to still work with
+#' most browsers.
+#'
+#' @section Miscellaneous Sass variables:
+#'
+#' Both Bootstrap 3 and 4 come with
+#'
+#' TODO: link to the variables files?
```
cpsievert, line 57: Clean up all the sections above

R/bs-dependencies.R
```diff
@@ -48,18 +36,18 @@
...
 #' # Bootswatch sketchy theme
-#' preview_button(bootstrap("sketchy"))
+#' preview_button(bs_dependencies("sketchy"))
 #' # Bootswatch solar theme with BS3 compatibility
-#' preview_button(bootstrap("solar@4+3"))
+#' preview_button(bs_dependencies("solar@4+3"))
 #'
-bootstrap <- function(
-  theme = bs_theme_get(),
+bs_dependencies <- function(
```
cpsievert, line 49: Let's make sure we deprcate `bootstrap()` and not completeltly remove it

R/bs-theme.R
```diff
@@ -1,3 +1,226 @@
...
+    base_font = base_font,
+    code_font = code_font,
+    heading_font = heading_font
+  )
+}
+
+#' @rdname bs_theme
+#' @param theme a [bs_theme()] object.
+#' @export
+bs_theme_update <- function(theme = bs_theme(), ..., bg = NULL, fg = NULL,
```
wch, line 168: Do we need a default value of `theme = bs_theme()`? Presumably if people call `bs_theme_update`, they have a thing they want to update.

R/bs-theme-preview.R
```diff
@@ -268,18 +256,21 @@ run_with_themer <- function(appDir = getwd(), ...) {
...
     stop(call. = FALSE, "bootstraplib::bs_themer() must be called from within a ",
-      "Shiny server function")
+         "Shiny server function")
   }
   if (!identical("ok", session$ns("ok"))) {
     stop(call. = FALSE, "bootstraplib::bs_themer() must be called from within a ",
-      "top-level Shiny server function, not a Shiny module server function")
+         "top-level Shiny server function, not a Shiny module server function")
+  }
+  theme <- shiny::getShinyOption("bs_theme")
```
cpsievert, line 267: @wch can/should we do something else that's better?

R/bs-theme-layers.R
```diff
@@ -0,0 +1,153 @@
...
+#' @export
+bs_add_layers <- function(theme = bs_theme(), ...) {
+  theme <- assert_bs_theme(theme)
+  is_layer <- vapply(rlang::list2(...), inherits, logical(1), "sass_layer")
+  if (!any(is_layer)) {
+    stop("`...` must contain a `sass::sass_layer()` object(s)")
+  }
+  add_class(sass_layer_merge(theme, ...), "bs_theme")
+}
+
```
cpsievert, line 125: Why is there no `bs_global_add_layers()` here?

R/bs-theme-layers.R
```diff
@@ -0,0 +1,153 @@
...
+bs_add_rules <- function(theme = bs_theme(), rules) {
+  bs_add_layers(theme, sass_layer(rules = rules))
+}
+
+# TODO: do we want global versions of every theming function?
+#bs_global_add_rules <- function(rules) {
+#  theme <- assert_global_theme("bs_global_add_rules()")
+#  theme <- bs_add_rules(theme, ...)
+#  bs_global_set(theme)
+#}
```
cpsievert, line 139: I think we can punt on this as long as we have `bs_global_add_layers()`

R/bs-dependencies.R
cpsievert: Ah, yea, this is in `deprecated.R`

R/bs-theme.R
cpsievert: No, we don't -- I was thinking it might make for useful short-hand code (i.e., `bs_theme_update(primary = "red")`), but I'm maybe we should make it more explicit

wch: If they call `bs_theme_update(primary = "red")` and intend it to create a new theme to modify, they might as well call `bs_theme(primary = "red")`, right?

I think it would be a little surprising to create a new theme object automatically for `bs_theme_create`(). So I'd vote for removing the default value for `theme`.

cpsievert: Yea, agreed

R/bs-theme-preview.R
wch: Now that I look at this in more detail, I think it's OK, since this is so tied with Shiny's internals. But didn't the option get renamed to something else?

R/bs-dependencies.R
```diff
@@ -48,18 +36,18 @@
...
+  theme = bs_global_get(),
```
wch, line 50: Maybe this function shouldn't have a default for `theme`, but `bootstrap()` could have it default to `bs_global_get()`?

cpsievert: Yea, that makes sense, I'll do the same for `bs_sass()`

R/bs-dependencies.R
```diff
@@ -1,27 +1,15 @@
-#' Compile Bootstrap 4 (or 3) SASS with (optional) theming
+#' Compile Bootstrap SASS with (optional) theming
```
wch, line 1: Nit: I think the proper case is "Sass".

schloerke:
```suggestion
#' Compile Bootstrap Sass with (optional) theming
```

R/bs-theme-preview.R
cpsievert: Great, yea, I handled the rename in https://github.com/rstudio/bootstraplib/pull/114/commits/c9293e9b9f461933b4301ccbd395f16e7473aa5a

tools/precompile-css.R
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env Rscript
```
wch, line 1: This was already moved into yarn_install.R, so there's no need to bring it back.

wch (approved):
Other than some small things, looks good.

wch (approved):

PR #115, opened 2020-09-24
Copy js map file over if it exists

schloerke:
Fixes https://github.com/rstudio/bootstraplib/issues/69


R/bs_sass.R
```diff
@@ -119,7 +132,7 @@ bootstrap <- function(
       htmlDependency(
         name = "bootstrap",
         version = get_exact_version(version),
-        src = dirname(out_file),
+        src = out_file_dir,
         stylesheet = basename(out_file),
         script = basename(js),
```
cpsievert, line 137: Since we're already assuming this behavior, let's be explicit about it

```suggestion
  all_files = TRUE,
```

schloerke, line 137:
```suggestion
        script = basename(js),
        all_files = TRUE,

```

R/bs_sass.R
```diff
@@ -107,10 +107,23 @@ bootstrap <- function(
...
+    js_map_file <- paste0(js, ".map")
+    # if `version` is `"3"`, then `js` is `"bootstrap-sass/javascripts/bootstrap[.min].js"`,
+    ## which does not have a map file
+    ## so check for map file existence
+    if (file.exists(js_map_file)) {
+      file.copy(js_map_file, out_file_dir, overwrite = TRUE)
+    } else {
+      TRUE
+    }
+  })
```
schloerke, line 124: Instead, we should remove the sourcemap link in the javascript file when updating the bootstrap files in `./tools`.

If users want access to the non-minified files, they can use `minified = FALSE`.

cpsievert (approved):
Pls update `tools/README.md` accordingly, then LGTM

PR #120, opened 2020-10-02
More generalized approach to real-time theming

cpsievert:
Requires https://github.com/rstudio/shiny/pull/3085


R/bs-theme-preview.R
```diff
@@ -358,21 +362,10 @@ bs_themer <- function() {
...
-      session$sendCustomMessage("bootstraplib-themer-disable-bootstrap", list(disable = TRUE))
-      # Now, insert the newly compiled Bootstrap CSS
       shiny::insertUI(
-        "head", where = "beforeEnd",
-        ui = tags$style(
-          id = "bs-realtime-preview-styles",
-          htmltools::HTML(css)
-        )
+        "body", where = "beforeEnd",
+        ui = div(id = "bs-themer-dependencies", deps)
```
cpsievert, line 367:
```suggestion
        ui = tagList(deps)
```

R/bs-dependencies.R
```diff
@@ -116,14 +116,34 @@ bs_dependencies <- function(
...
+#' @param func a function that takes a [bs_theme()] object (or `theme_default`) as input and
+#' and returns an [htmlDependency()] (or a list of them).
+#' @param theme_default In the case that `func` wants to be executed outside of a Shiny
+#' runtime (e.g., in an R Markdown doc), an expression that yields a value to provide to `func`.
+#' @param env environment for evaluation the expression provided to `theme_default`.
+#' @export
+bs_runtime_dependencies <- function(func, theme_default = bs_global_get(), envir = parent.env()) {
+  theme_default <- substitute(theme_default)
+  tagFunction(function() {
+    if (is_available("shiny") && shiny::isRunning()) {
```
wch, line 137: It may be better to check if shiny is loaded -- otherwise, this code will cause shiny to be loaded even if this is being used for an rmarkdown document.

R/bs-dependencies.R
```diff
@@ -124,39 +124,133 @@ bs_dependencies <- function(
...
+#'   if (!is_bs_theme(theme)) {
+#'     return(
+#'       htmlDependency(
+#'         name = "mywidget-css",
+#'         version = myWidgetVersion,
+#'         src = system.file(package = "mypackage", "css"),
+#'         stylesheet = "mywidget.css"
+#'       )
+#'     )
+#'   }
```
cpsievert, line 195: I was thinking it might make sense for `bs_dependency_dynamic()` to only ever call `func` if `theme` is a `bs_theme()`. I don't feel that strongly about it, but it seems like it might be more intuitive to keep your static vs dynamic dependency logic separated?

cpsievert: Ohh, thinking about it more, I suppose there could be a situation where an update to a `NULL` theme should trigger a restyle, so I think this is the right approach

wch: I was initially headed in the direction of always including the static CSS, and having this part either return `NULL` (if `theme` is not a `bs_theme`) or the compiled Sass, with the idea that the compiled Sass would override the static CSS, but there were two problems:

* The `version` for the compiled Sass wouldn't cause it to override the static CSS
* The code that that authors would write was really convoluted. I can provide an example a bit later.

R/bs-dependencies.R
```diff
@@ -125,18 +125,125 @@ bs_dependencies <- function(
...
+  )
+  do.call(htmlDependency, dep_args)
+}
+
+
+#' Note that `func` should not be created in a closure -- for Shiny to
+#' deduplicate the html dependencies, it needs to be able to tell that the
+#' `func` from a previous invocation is `identical()` to the `func` in later
+#' invocations. In order to that, the
+#'
```
cpsievert, line 163:
```suggestion
```

R/bs-dependencies.R
```diff
@@ -125,18 +125,125 @@ bs_dependencies <- function(
...
+#'   )
+#' }
+#'
+#' # Note that myWidgetDepdency is not defined inside of myWidget. This is so
+#' # that, if `myWidget()` is called multiple times, Shiny can tell that the
+#' # function objects are identical and deduplicate them.
+#' # TODO: Maybe Shiny should check for identical outputs, and warn in that case
+#' # that a dependency function isn't being deduped.
+#' # TODO: Check if/how browser reloads mywidget/mywidget.css even if the URL
+#' # is the same a second time.
```
cpsievert, line 217:
```suggestion
```

wch (approved):

PR #135, opened 2020-10-27
Fixes for real-time themer

cpsievert:


R/utils.R
```diff
@@ -64,13 +64,14 @@ names2 <- function(x) {
...
-    names(x)[names(x) %in% nm] <- names(nm)
+#' rename(list(a=1, b=3, c=4, d=1), b="z", a="y", f="w"),
+#' #> list(y = 1, z = 3)
+rename <- function(x, ...) {
+  defs <- list(...)
+  matches <- intersect(names(x), names(defs))
+  for (i in seq_along(matches)) {
+    m <- matches[i]
+    names(x)[names(x) %in% m] <- defs[m]
   }
```
wch, line 75: I can't abide by the looping and modification in place. :)

This should be faster. It's adapted from `plyr::rename` and `plyr::mapvalues`:

```suggestion
  map <- match(names(x), names(defs))
  mapidxNA <- is.na(map)
  names(x)[!mapidxNA] <- defs[map[!mapidxNA]]
```

R/utils.R
```diff
@@ -64,13 +64,14 @@ names2 <- function(x) {
...
 #' @examples
-#' rename(list(a = 1, b = 3), c(z = "b", y = "a"))
-rename <- function(x, nms) {
-  nms_idx <- match(nms, names(x))
-  nms_idx <- nms_idx[!is.na(nms_idx)]
-  for (i in nms_idx) {
-    nm <- nms[i]
-    names(x)[names(x) %in% nm] <- names(nm)
+#' rename(list(a=1, b=3, c=4, d=1), b="z", a="y", f="w"),
+#' #> list(y = 1, z = 3)
```
wch, line 68: The original had a trailing comm and the output didn't match with actual behavior.

This version also has two instances of `a`, to demonstrate that multiple instances will be renamed.

```suggestion
#' rename2(list(a=1, b=3, c=4, a=2), b="z", f="w", a="y")
#' #> list(y = 1, z = 3, c = 4, y = 2)
```

tests/testthat/test-global.R
```diff
@@ -60,7 +60,7 @@ test_that("Theme adding works as intended", {
 
 test_that("rename works as intended", {
   expect_identical(
-    rename(list(a=1, b=3, c=4, d=1), c(z="b",y="a",w="f")),
+    rename(list(a=1, b=3, c=4, d=1), b = "z", a = "y", f = "w"),
     list(y=1, z=3, c=4, d=1)
```
wch, line 64: Slight change in the test, for the reasons mentioned earlier.

```suggestion
    rename2(list(a=1, b=3, c=4, a=2), b="z", f="w", a="y"),
    list(y = 1, z = 3, c = 4, y = 2)
```

wch (approved):

PR #139, opened 2020-10-28
Web font importing

### LINKED ISSUE #112: Make bs_theme_fonts() 'just work' for Google Fonts
By attaching an HTML dependency to the theme that imports the font

```r
bs_theme_fonts <- function(..., google_import = c("href", "file", "none")) {
  ...
}
```

cpsievert:
(Closes #112)

This PR adds three functions for making it easier to import font resources when setting new font defaults, namely `font_google()` for (local or remote) Google Fonts, `font_link()` for simple remote fonts, and `font_face()` for a more direct interface to CSS `@font-face`. This also adds the ability to enter a Google Font into the real-time themer and have it emit the proper code

```r
theme <- bs_theme(
  base_font = font_google("Fira Sans"),
  code_font = font_google("Fira Code"),
  heading_font = font_google("Fredoka One")
)
bs_theme_preview(theme)
```


R/bs-theme.R
```diff
@@ -43,42 +43,35 @@
...
+#'  However, specifying multiple "fallback" font families is recommended,
+#'  especially when relying on remote and/or system fonts being available, for
+#'  instance. Fallback fonts are useful not only for handling missing fonts, but
+#'  also for handling a Flash of Invisible Text (FOIT) which can be quite
+#'  noticeable with remote web fonts on a slow internet connection.
 #'
 #'  ```
-#'  bs_theme(base_font = "Source Sans Pro, sans-serif")
-#'  ```
+#'  bs_theme(base_font = list(gfont("Pacifico", local = FALSE), "Roboto", "sans-serif")
```
cpsievert, line 73: Update this

R/fonts.R
```diff
@@ -0,0 +1,378 @@
...
+    "\n}"
+  )
+}
+
+
+font_dep_link <- function(x, version) {
+  htmltools::htmlDependency(
+    font_dep_name(x), version,
+    head = format(tags$link(
+      href = httpuv::encodeURI(font_url(x)),
```
cpsievert, line 195: Use `URLencode()` instead

R/bs-dependencies.R
```diff
@@ -111,7 +111,7 @@ bs_theme_dependencies <- function(
     warning("Failed to copy over bootstrap's javascript files into the htmlDependency() directory.")
   }
 
-  c(
+  htmltools::resolveDependencies(c(
```
wch, line 114: Why is this changed to `resolveDependencies`?

R/fonts.R
```diff
@@ -0,0 +1,442 @@
...
+        "https://fonts.googleapis.com/css2?family=",
+        family, axis_rng, "&display=", display
+      )
+    )),
+    class = "bs_font_google"
+  )
+}
+
+is_font_object <- function(x) {
+  is_font_face(x) || is_font_link(x) || is_font_google(x)
```
wch, line 182: What about just having a second class for all these objects, named `bs_font`?

R/bs-dependencies.R
cpsievert: IIRC, it was to handle the new `tagFunction()` that comes with `bs_font_dependencies()`, but come to think of it, maybe `setCurrentTheme()` should be resolving dependencies?

R/fonts.R
cpsievert: Yea, that seems like a good idea

cpsievert: I do need to be able to differentiate them though

R/bs-dependencies.R
cpsievert: After thinking about this a bit more, this seems fine for now, but we should probably do this inside `setCurrentTheme()` since, in some edge cases, folks might want to call `bs_theme_dependencies()` directly without resolving the dependencies straight away.

wch (approved):

PR #149, opened 2020-11-04
Bump bootstrap & bootswatch to 4.5.3, closes #147

### LINKED ISSUE #147: Update to 4.5.3
Both core and bootswatch

cpsievert:
The only "real" changes here are the bumping of versions in `inst/package.json` and accounting for the new use of https://caniuse.com/css-color-adjust in `tools/yarn_install.R `


schloerke (approved):
I'd be happy to add a check for the precompiled to auto push if you'd like

PR #158, opened 2020-11-11
Fix google_font(local = TRUE) caching logic, closes #157

### LINKED ISSUE #157: font_google(local = TRUE) imports remote instead of local files
Minimal example:

```r
library(shiny)
library(bslib)

ui <- fluidPage(
  theme = bs_theme(base_font = font_google("Pacifico")),
  "Hello"
)

shinyApp(ui, function(input, output) {})
```

Note that the font renders ok, but the `font.css` for the relevant dependency  contains a `url()` pointing to a remote source -- it should be pointing the local file!

cpsievert:
In addition to closing #157, this also accounts for a cache hit on the CSS file, but a cache miss on the relevant font files


cpsievert:
Merging without review since folks are gonna build up bad caches without this change

Issue #110, opened 2020-09-18
Provide a key value store of available versions


cpsievert:
Maybe it should even be exported?

https://github.com/rstudio/bootstraplib/pull/109#discussion_r491023391

PR #160, opened 2020-11-11
Get rid of version = 4+3 and make version = 4 include compatibility

### LINKED ISSUE #153: Change version = 4+3 to version = 4 to avoid confusion
Barely anyone will want Bootstrap 4 without the compatibility layer, so make that something like `version = "4strict"`?

cpsievert:
Closes #153, #159, and #110.

This PR deprecates `version = 4+3` and changes `version = 4` to now include the compatibility layer. It also adds a `bs_remove()` which can be used to remove that compatibility layer if need be: `bs_remove(bs_theme(version = 4), "bs3compat")`. `bs_remove()` can also be used to remove Bootstrap rules (e.g., `bs_remove(bs_theme(version = 4), "_print")`)


R/bootswatch.R
```diff
@@ -13,44 +13,39 @@ bootswatch_themes <- function(version = version_default(), full_path = FALSE) {
...
+  version <- grep("^bs_version_", class(theme), value = TRUE)
+  sub("^bs_version_", "", version)
 }
 
 
 bootswatch_dist <- function(version) {
-  if (version %in% "3") {
+  switch_version(
+    version,
+    lib_file("bootswatch", "dist"),
```
cpsievert, line 39: specify arg names in switch_version()

R/bootswatch.R
```diff
@@ -13,44 +13,39 @@ bootswatch_themes <- function(version = version_default(), full_path = FALSE) {
...
-  }
+  )
 }
 
 bootswatch_theme_resolve <- function(bootswatch, version) {
   if (is.null(bootswatch)) return("bootstrap")
   # because rmarkdown
   if (bootswatch %in% c("default", "bootstrap", "")) return("bootstrap")
-  if (version %in% c("4", "4+3")) {
+  if (version %in% "4") {
```
cpsievert, line 48: Let's avoid this as much as possible and use `switch_version()` instead

R/bs-dependencies.R
```diff
@@ -276,20 +272,6 @@ bs_dependency_defer <- function(func) {
 as_bs_theme <- function(theme) {
   if (is_bs_theme(theme)) return(theme)
 
-  # Allow users to do something like
-  # bs_theme_dependencies(theme = sass_bundle(bs_global_get(), my_layer()))
-  if (is_sass_bundle(theme) || inherits(theme, "sass_layer")) {
-    theme <- add_class(
```
cpsievert, line 282: Let's throw here just in case

R/bs-theme-layers.R
```diff
@@ -100,12 +100,42 @@ ensure_default_flag <- function(x) {
...
+#'   provided (e.g., a list, character vector, [sass::sass_file()], etc)
 #' @export
 bs_add_rules <- function(theme, rules) {
   bs_bundle(theme, sass_layer(rules = rules))
 }
 
+#' @describeIn bs_bundle Remove Sass rules.
+#' @param file a character vector referencing filenames from the `libname`
+#' @param libname a character string referencing the library of interest.
+#' Possible value include `'bootstrap'`, `'bootswatch'`, `'bs3compat'`, and `'accessibility'`
```
cpsievert, line 113:
```suggestion
```

R/bs-theme-layers.R
```diff
@@ -100,12 +100,42 @@ ensure_default_flag <- function(x) {
...
+  bundle_ids <- names(theme$layers)
+  if (!length(ids)) {
+    stop(
+      "Provide one or more of the following `ids`: ",
+      paste(ids, collapse = ", ")
+    )
+  }
+  missing_ids <- setdiff(ids, bundle_ids)
+  if (length(missing_ids)) {
+    stop(
```
cpsievert, line 129:
```suggestion
    warning(
```

R/bs-theme-update.R
```diff
@@ -12,7 +12,6 @@ bs_base_colors <- function(theme, bg = NULL, fg = NULL) {
   args <- list(bg = bg, fg = fg)
   args <- validate_and_normalize_colors(args)
   funcs <- list(
```
cpsievert, line 14: Should we do `switch_version()` here?

R/utils.R
```diff
@@ -1,14 +1,22 @@
...
+  match.arg(version, c("4", "3"))
+}
+
+switch_version <- function(version, four, three) {
+  switch(
+    version_resolve(version),
+    `4` = four, `3` = three,
+    stop("Didn't recognize Bootstrap version: ", version, call. = FALSE)
+  )
 }
```
cpsievert, line 16:
```suggestion
switch_version <- function(version, four = default, three = default, default = NULL) {
  switch(
    version_resolve(version),
    `4` = four, `3` = three,
    stop("Didn't recognize Bootstrap version: ", version, call. = FALSE)
  )
}
```

README.Rmd
```diff
@@ -151,6 +152,25 @@ Moreover, when `theme` depicts a `bs_theme()`, you may modify the (global) `them
...
+  )
+```
+
+To learn more, see the [theming article](https://rstudio.github.io/bslib/articles/theming.html).
+
+### Reducing theme size
+
+By default, `bs_theme()` includes many CSS rules, including styles for every Bootstrap Component, many of which aren't need for most projects. Simply apply `bs_remove_rules()`  `my_theme` to 
+
+To get a "recommended" set of rules for "core" Shiny & R Markdown UI just do `bs_remove_rules(my_theme)`, but you can also target specific rules that you want (e.g., `bs_remove_rules(my_theme, c("carousel", "pagination"))`).
```
cpsievert, line 172: Need to update this section (should it even be mentioned in README?)

R/bs-remove.R
```diff
@@ -0,0 +1,65 @@
...
+bs_remove <- function(theme, ids = character(0)) {
+  ids <- retain_valid_ids(theme, ids)
+  sass_bundle_remove(theme, ids)
+}
+
+#' @rdname bs_remove
+#' @param include_unnamed whether or not to include unnamed [sass::sass_layer()]s
+#' (e.g., Bootstrap Sass variables, functions, and mixins).
+#' @export
+bs_retrieve <- function(theme, ids = character(0), include_unnamed = TRUE) {
```
schloerke, line 29: My take would be to set `include_unnamed = FALSE` as I would be looking for a particular set of ids

```suggestion
bs_retrieve <- function(theme, ids = character(0), include_unnamed = FALSE) {
```

R/bs-theme.R
```diff
@@ -397,40 +387,38 @@ bootswatch_bundle <- function(bootswatch, version) {
...
-      bootswatch_sass_file(bootswatch, "variables", version)
-    ),
-    rules = list(
-      bootswatch_sass_file(bootswatch, "bootswatch", version),
-      # For some reason sketchy sets .dropdown-menu{overflow: hidden}
-      # but this prevents .dropdown-submenu from working properly
-      # https://github.com/rstudio/bootscss/blob/023d455/inst/node_modules/bootswatch/dist/sketchy/_bootswatch.scss#L204
-      if (identical(bootswatch, "sketchy")) as_sass(".dropdown-menu{ overflow: inherit; }") else ""
+  sass_bundle(
+    bootswatch := sass_layer(
```
schloerke, line 391: Does the left side need a `!!`?

```suggestion
    !!bootswatch := sass_layer(
```

tests/testthat/test-zzzz-bs-sass.R
```diff
@@ -3,7 +3,7 @@ local_disable_cache()
 test_that("Can access the sass behind all versions and Bootswatch themes", {
   skip_on_cran()
 
-  versions <- c("4+3", "3", "4")
+  versions <- c("3", "4")
   for (version in versions) {
```
schloerke, line 7:
```suggestion
  for (version in versions()) {
```

tools/yarn_install.R
```diff
@@ -327,7 +327,7 @@ for (patch in list.files(patch_dir, full.names = TRUE)) {
 library(bslib)
 
 # The versions of Bootstrap to precompile
-versions <- c("4", "4+3", "3")
+versions <- c("4", "3")
```
schloerke, line 330: Remove this precompile setup and use `versions()` in the `lapply` below.

schloerke (approved):
LGTM pending small changes

R/bs-remove.R
cpsievert: The thing is, from a user perspective, it doesn't make much sense to retrieve `rules` without the `defaults` and `declarations`. For example, this would error out:

https://github.com/rstudio/bslib/pull/160/files#diff-c3c3731e2b8667ea34c6a7c9625e6e755ae8ae02ff1527d418400f1052e71641R17

So, I think `TRUE` is a better default because it's "safer" in some sense

R/bs-theme.R
cpsievert: No, this is intentional so that you can do `bs_remove(bs_theme(), "bootswatch")` and have it work regardless of the theme

PR #162, opened 2020-11-17
Remove rlang walrus for static key

schloerke:
Ref: https://github.com/rstudio/bslib/pull/160#discussion_r524554344

> No, this is intentional so that you can do `bs_remove(bs_theme(), "bootswatch")` and have it work regardless of the theme

Then it should be static with no walrus sign.


cpsievert:
Oops, thanks!

PR #177, opened 2020-11-25
Hosted version of bs_theme_preview()

cpsievert:
Makes a version of `bs_theme_preview()` available at https://testing-apps.shinyapps.io/themer-demo/

This should be merged after https://github.com/rstudio/thematic/pull/68


schloerke (approved):
LGTM. 

If you can add in an auto deploy using GHA secrets and calling your deploy script which listens to the secrets, then this app would always be up to date. Ex: https://github.com/rstudio/learnr/blob/7eb2b298699733a9a64d9c15a728b0ca6a4d68d5/.github/workflows/R-CMD-check.yaml#L102-L158

cpsievert:
Merging before confirming that deploy was successful since GHA is down

PR #180, opened 2020-12-02
htmltools::parseCssColors now requires character vector input

cpsievert:
Follow up to https://github.com/rstudio/htmltools/pull/194


cpsievert:
Merging before review since QA needs this fix for testing

PR #185, opened 2020-12-11
Proper scoping of color contrasting parameters

cpsievert:
Provides a fix for `bs_theme_preview(bs_theme(bootswatch = "solar"))`

### DIFF
--- a/inst/sass-utils/color-contrast.scss
+++ b/inst/sass-utils/color-contrast.scss
@@ -9,6 +9,7 @@ $_luminance-list: .0008 .001 .0011 .0013 .0015 .0017 .002 .0022 .0025 .0027 .003
 @function color-contrast($background) {
   // These variables should be defined in _variables.scss, but we also
   // define them here so that 3rd party libs can use if they want
+  // without polluting the global namespace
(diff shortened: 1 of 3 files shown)


inst/sass-utils/color-contrast.scss
```diff
@@ -6,14 +6,15 @@
...
-// define them here so that 3rd party libs can use if they want
-$black: #000 !default;
-$white: #fff !default;
-$color-contrast-dark: $black !default;
-$color-contrast-light: $white !default;
-$min-contrast-ratio: 3 !default;
-@function color-contrast($background, $color-contrast-dark: $color-contrast-dark, $color-contrast-light: $color-contrast-light, $min-contrast-ratio: $min-contrast-ratio) {
+@function color-contrast($background) {
+  // These variables should be defined in _variables.scss, but we also
+  // define them here so that 3rd party libs can use if they want
```
schloerke, line 11:
```suggestion
  // define them here so that 3rd party libs can use if they want
  // without polluting the global namespace
```

schloerke (approved):
LGTM given a lil more context on the comment

PR #186, opened 2020-12-11
.shiny-text-output no longer extends .well

cpsievert:
Won't be necessary after https://github.com/rstudio/shiny/pull/3203


wch (approved):

wch:
Looks like the css file needs to be precompiled?

PR #199, opened 2020-12-17
Different inline code styling for BS4

### LINKED ISSUE #181: Add $code-bg to BS4
BS3 has it, and it's probably better to keep it so that by default code styling is similar, but folks would still have abaility to turn it off with `$code-bg: null`

cpsievert:
Closes #181


hadley:
Could you screenshot a rendered example?

cpsievert:
```yaml
html_document:
    theme: 
      version: 4
```

Screen Shot 2020-12-18 at 9 45 07 AM

```yaml
theme:
      bg: "#121212"
      fg: "#E4E4E4"
```

Screen Shot 2020-12-18 at 9 46 48 AM

cpsievert:
And now with no accent color by default:

Screen Shot 2020-12-18 at 12 39 11 PM


Screen Shot 2020-12-18 at 12 40 15 PM

hadley:
Yeah, I think that looks a lot better.

nstrayer:
I almost wonder if ideally the text color should be adjusted slightly to maintain constant contrast levels between the text and background. With the grey background, the code feels a little more fatiguing to read than the normal text. Not sure how easy that is and may require some JS to do...

nstrayer:
A quick riff on this: let `L(color)` = luminance value of color.
The contrast ratio as seen in [this article](https://medium.muz.li/the-science-of-color-contrast-an-expert-designers-guide-33e84c41d156) is:
`(L(bg) + 0.05) / (L(txt) + 0.05)` (assuming light-mode).
This means that the luminance of the color for the code text is 
`L(code_txt) = [ ((L(txt) + 0.05) / (L(bg) + 0.05) ) * (L(code_bg) + 0.05) ] - 0.05`. 
Just adding or subtracting uniform values from the plain text RGB until the [W3 luminance calculation](https://www.w3.org/WAI/GL/wiki/Relative_luminance) sits at `L(code_txt)` should get us the optimal color...

... or we could just skip all the math and bump the luminance up or down a set amount.

cpsievert:
Good point, here's the latest with slightly higher contrast:

Screen Shot 2020-12-18 at 1 53 10 PM

Screen Shot 2020-12-18 at 1 53 31 PM

tools/patches/009-bs4-code-styles.patch
```diff
@@ -0,0 +1,53 @@
...
+--- a/inst/lib/bootstrap/scss/_variables.scss
++++ b/inst/lib/bootstrap/scss/_variables.scss
+@@ -1117,7 +1117,11 @@ $close-text-shadow:                 0 1px 0 $white !default;
+ // Code
+ 
+ $code-font-size:                    87.5% !default;
+-$code-color:                        $pink !default;
++// Amount intentionally matches mixture amount of shiny::inputPanel()
++// https://github.com/rstudio/shiny/blob/4eeb4a1/inst/www/shared/shiny_scss/bootstrap.scss#L25
++$code-bg:                           mix($body-bg, $body-color, 96%)!default;
```
cpsievert, line 37:
```suggestion
+$code-bg:                           mix($body-bg, $body-color, 96%) !default;
```

inst/lib/bootstrap/scss/_variables.scss
```diff
@@ -1117,15 +1117,21 @@ $close-text-shadow:                 0 1px 0 $white !default;
 // Code
 
 $code-font-size:                    87.5% !default;
-$code-color:                        $pink !default;
+// Amount intentionally matches mixture amount of shiny::inputPanel()
+// https://github.com/rstudio/shiny/blob/4eeb4a1/inst/www/shared/shiny_scss/bootstrap.scss#L25
+$code-bg:                           mix($body-bg, $body-color, 96%)!default;
```
cpsievert, line 1122:
```suggestion
$code-bg:                           mix($body-bg, $body-color, 96%) !default;
```

nstrayer (approved):
Looks nice to me!

PR #224, opened 2021-01-14
Fixes and changes to bootstrap accessibility plugin

cpsievert:
This fixes the same bug as https://github.com/rstudio/shiny/pull/3256 (for `theme = bslib::bs_theme(version = 3)`) and also changes `$mouse-outline-style` to `none` so that you must opt-in with something like this:

https://github.com/paypal/bootstrap-accessibility-plugin has several issues:

(1) [It assumes that the page is already loaded before it runs](https://github.com/paypal/bootstrap-accessibility-plugin/blob/fbbf870eafc1ee5d4547fabbeea71778ddbe2166/src/js/tab.js#L4-L12), so the aria labels aren't added on page load to something like this:

```{r}
library(shiny)
ui <- fluidPage(
  theme = bslib::bs_theme(version = 3),
  tabsetPanel(
    tabPanel("A", "a"),
    tabPanel("B", "b")
  )
)
shinyApp(ui, server)
```

(2) It always provides outlines for mouse interactions, which is quite a deviation in visual style, and probably annoying most users, so we've made these mouse outlines an opt-in feature via a Sass variable called `$mouse-outline-style`:

```r
library(shiny)
ui <- fluidPage(
  theme = bslib::bs_theme(version = 3, "mouse-outline-style" = "auto"),
  tabsetPanel(
    tabPanel("A", "a"),
    tabPanel("B", "b")
  )
)
shinyApp(ui, server)
```

(3) As a consequence of opt-ing out of mouse-based outlines, keyboard navigation suffers on Chrome, so we've decided to enable `:focus-visible` outlines by default, so keyboard navigation is supported on browsers that support it.

```r
library(shiny)
ui <- fluidPage(
  theme = bslib::bs_theme(version = 3, "keyboard-outline-color" = "red"),
  tabsetPanel(
    tabPanel("A", "a"),
    tabPanel("B", "b")
  )
)
shinyApp(ui, server)
```


tools/yarn_install.R
```diff
@@ -336,6 +337,44 @@ for (patch in list.files(patch_dir, full.names = TRUE)) {
...
 }
 
+# ----------------------------------------------------------------------
+# Apply minification to patched files
+# ----------------------------------------------------------------------
+
+withr::with_dir("inst", {
+  local({
+    # install all deps
+    system("yarn install")
```
wch, line 347: Why no `--production` here if it's added above?

tools/yarn_install.R
```diff
@@ -11,7 +11,8 @@ if (Sys.which("yarn") == "") {
   stop("The yarn CLI must be installed and in your PATH")
 }
 
-withr::with_dir("inst", system("yarn install"))
+# only install the direct deps
+withr::with_dir("inst", system("yarn install --production"))
```
wch, line 15: Are we sure that this is the right way to go?

inst/lib/bs-a11y-p/src/sass/partials/_mixins.scss
```diff
@@ -1,3 +1,6 @@
 @mixin outline($size) {
```
wch, line 1: Can you add a comment that says that `outline` typically isn't visible?

wch: Or possibly rename `outline` to something that makes more sense.

wch (approved):

tools/yarn_install.R
cpsievert: This makes it so we can use `devDependencies` without having to clean-up parcel

tools/yarn_install.R
cpsievert: So that `yarn parcel` is available (it isn't with `--production`)

schloerke (approved):

PR #226, opened 2021-01-15
Use `tools::R_user_dir()` when possible

wch:
This PR addresses this comment from CRAN:

```
Also, checking creates and leaves behind
   ~/.cache/R-bslib

in violation of the CRAN Policy's

   Packages should not write in the user's home filespace (including
   clipboards), nor anywhere else on the file system apart from the R
   session's temporary directory (or during installation in the location
   pointed to by TMPDIR: and such usage should be cleaned up). Installing
   into the system's R installation (e.g., scripts to its bin directory)
   is not allowed.

   Limited exceptions may be allowed in interactive sessions if the
   package obtains confirmation from the user.

   For R version 4.0 or later (hence a version dependency is required or
   only conditional use is possible), packages may store user-specific
   data, configuration and cache files in their respective user
   directories obtained from tools::R_user_dir(), provided that by
   default sizes are kept as small as possible and the contents are
   actively managed (including removing outdated material).

Can you please change as necessary, ideally using the
tools::R_user_dir() mechanism where available?
```

When re-submitting to CRAN, it would be a good idea to mention that the behavior is fixed on R 4.0, but unchanged for older versions of R.

### DIFF
--- a/R/fonts.R
+++ b/R/fonts.R
@@ -381,6 +381,18 @@ hash <- function(x) {
   digest::digest(x, algo = "spookyhash")
 }
 
+
+find_cache_dir <- function(pkg) {
+  # In R 4.0 and above, CRAN wants us to use the new tools::R_user_dir().
+  # If not present, fall back to rappdirs::user_cache_dir().
+  R_user_dir <- .getNamespace('tools')$R_user_dir
+  if (!is.null(R_user_dir)) {
+    R_user_dir(pkg)
+  } else {
+    rappdirs::user_cache_dir(paste0("R-", pkg))
+  }
+}
+
 # Same idea as sass::sass_cache_context_dir, but different dir
 cache_context_dir <- function(pkg = "bslib") {
   tryCatch(
--- a/R/fonts.R
+++ b/R/fonts.R
@@ -385,7 +385,7 @@ hash <- function(x) {
 find_cache_dir <- function(pkg) {
   # In R 4.0 and above, CRAN wants us to use the new tools::R_user_dir().
   # If not present, fall back to rappdirs::user_cache_dir().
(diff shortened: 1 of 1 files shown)


R/fonts.R
```diff
@@ -381,12 +381,24 @@ hash <- function(x) {
   digest::digest(x, algo = "spookyhash")
 }
 
+
+find_cache_dir <- function(pkg) {
+  # In R 4.0 and above, CRAN wants us to use the new tools::R_user_dir().
+  # If not present, fall back to rappdirs::user_cache_dir().
+  R_user_dir <- .getNamespace('tools')$R_user_dir
```
cpsievert, line 388: Why not `asNamespace()` -- is this faster?

schloerke: `asNamespace(x)` wraps around `getNamespace(x)`, which does a little more than `.getNamespace(x)`. Maybe use `getNamespace("tools")` to be safer?

wch: For future reference: `getNamespace()` is better than `.getNamespace()`, since the former will cause `tools` to be loaded if it wasn't already. If `Rscript` is used (instead of `R`), `tools` won't be loaded by default:


```
$ Rscript -e "loadedNamespaces()"
[1] "compiler"  "graphics"  "utils"     "grDevices" "stats"     "datasets" 
[7] "methods"   "base"     
```


However, if `getNamespace('tools')` is called, then it will cause the package to load:
```
$ Rscript -e "getNamespace('tools'); loadedNamespaces()"
<environment: namespace:tools>
[1] "compiler"  "graphics"  "tools"     "utils"     "grDevices" "stats"    
[7] "datasets"  "methods"   "base"     
```

PR #238, opened 2021-01-25
Add new vignette that contains table of all sass variables

### LINKED ISSUE #189: Provide a table of Bootstrap Sass variables
So it's more approachable than linking to Sass files. Broken down by version in a tabset in a pkgdown article?

nstrayer:
This PR contains a new vignette/pkgdown article that has an interactive table of all of the possible sass variables present in Bootstrap 4. 

Closes #189


vignettes/available-sass-variables.Rmd
```diff
@@ -0,0 +1,249 @@
...
+  cursor: pointer;
+}
+```
+
+```{r variables_df}
+# BS3
+# sass_variables_loc <- "https://raw.githubusercontent.com/rstudio/bslib/master/inst/lib/bs-sass/assets/stylesheets/bootstrap/_variables.scss"
+
+# BS4
+sass_variables_loc <- "https://raw.githubusercontent.com/rstudio/bslib/master/inst/lib/bs/scss/_variables.scss"
```
nstrayer, line 56: This should probably be switched to using the current file in the repo itself.

vignettes/available-sass-variables.Rmd
```diff
@@ -0,0 +1,249 @@
...
+  "paragraph-margin-bottom" = "2rem"
+) %>% 
+  # When referring to other variables, prefix with dollar-sign
+  bs_add_variables("dark" = "$gray-900", .where = "declarations")
+```
+
+For more details on this see the `sass` variables section of the [theming article.](https://rstudio.github.io/bslib/articles/theming.html#sass-variables-1)
+
+
+## All customizable variables
```
nstrayer, line 177: Should specify that this is only for BS4.

vignettes/available-sass-variables.Rmd
```diff
@@ -0,0 +1,251 @@
...
+}
+```
+
+```{r variables_df}
+# BS3
+# Not using BS3 now because the comment templates are different so a whole new scaping pipeline is needed
+# sass_variables_loc <- "https://raw.githubusercontent.com/rstudio/bslib/master/inst/lib/bs-sass/assets/stylesheets/bootstrap/_variables.scss"
+
+# BS4
+sass_variables_loc <- "../inst/lib/bs/scss/_variables.scss"
```
cpsievert, line 57:
```suggestion
sass_variables_loc <- rprojroot::find_package_root_file("inst/lib/bs/scss/_variables.scss")
```

_pkgdown.yml
```diff
@@ -19,6 +19,8 @@ navbar:
   left:
   - text: Custom themes
     href: articles/theming.html
+  - text: Sass variables
+    href: articles/available-sass-variables.html
```
cpsievert, line 23: I think I'd prefer to default to calling these "theming variables", and then in the longer form writing, explain that theming vars are actually Sass variables (but only when Sass becomes relevant)

vignettes/available-sass-variables.Rmd
```diff
@@ -0,0 +1,251 @@
...
+```
+
+```{r setup}
+library(dplyr)
+library(readr)
+library(stringr)
+library(purrr)
+library(glue)
+library(bslib)
+```
```
cpsievert, line 31: I think we'll probably need to add these to https://github.com/rstudio/bslib/blob/master/.github/workflows/pkgdown.yaml 

(which redeploys the pkgdown site everytime we push to master)

vignettes/available-sass-variables.Rmd
nstrayer: Oh fancy and useful.

_pkgdown.yml
nstrayer: That's much nicer, I agree.

vignettes/available-sass-variables.Rmd
```diff
@@ -0,0 +1,251 @@
...
+
+Bootstrap uses `sass` to generate its final CSS. While a large portion of the customization you may want to do to your theme is easily done with the common arguments in `bs_theme()`, sometimes you may want finer-grain control. To enable this, `bslib` let's you change any of the variables used in the Bootstrap `sass` code. For instance, if you want to adjust the padding amount after each paragraph, that is available with the variable `paragraph-margin-bottom`, if we wanted to double the default amount (`1rem`) we could do so by simply passing the name of the variable to our `bs_theme()` call...
+
+```{r, eval = FALSE, echo = TRUE}
+theme <- bs_theme(
+  ...,
+  # Double space given after each paragraph
+  "paragraph-margin-bottom" = "2rem"
+)
+```
```
cpsievert: You might be able to convince me otherwise, but I might prefer to not mention `Sass` right away (and instead give a shorter intro), then go into more details below the table

vignettes/available-sass-variables.Rmd
nstrayer: I think I did this correctly here: https://github.com/rstudio/bslib/pull/238/commits/14094cee7df7dd3178eeb72f67c7f7dcdc494628

_pkgdown.yml
nstrayer: https://github.com/rstudio/bslib/pull/238/commits/bbcfb1b4b04286f5890c35e61a966a7a9f58b123

vignettes/available-sass-variables.Rmd
nstrayer: I agree. It felt like the intro got a bit out of hand. I think since this is nicely covered in the other vignette it can be very-much trimmed down.

vignettes/available-sass-variables.Rmd
```diff
@@ -0,0 +1,251 @@
...
+```{r, eval = FALSE, echo = TRUE}
+theme <- bs_theme(
+  ...,
+  "paragraph-margin-bottom" = "2rem"
+) %>% 
+  # When referring to other variables, prefix with dollar-sign
+  bs_add_variables("dark" = "$gray-900", .where = "declarations")
+```
+
+For more details on this see the `sass` variables section of the [theming article.](https://rstudio.github.io/bslib/articles/theming.html#sass-variables-1)
```
cpsievert: I think this should probably go below the table and we should consider removing the `$` in the variable column of the table (then, hopefully, users won't be so quick to wonder what the `$` is all about).

We'll have to, of course, explain why things like `$black` appear in the value column, as well as explain how to set a variable to default to another variable (e.g., `$dark: $gray-900`), but it seems like this should be more an after-thought (plus we could link to the other article, which also explains this)

vignettes/available-sass-variables.Rmd
```diff
@@ -0,0 +1,251 @@
...
+
+
+## All customizable variables
+
+There are a _lot_ of these `sass` variables available for adjusting. Below is a table that shows all of them, along with basic comments on variables functionality as provided in the original Bootstrap code. For more details on how customization of these variables works, see the function `bs_add_variables()`.
+
+_Note: these variables are specifically for Bootstrap 4 and many won't work for Bootstrap 3._
```
cpsievert: This should be more prominent (put above or in the table?)

.github/workflows/pkgdown.yaml
```diff
@@ -57,6 +57,11 @@ jobs:
         shell: Rscript {0}
         run: |
           pak::pkg_install("pkgdown", dependencies = TRUE)
+          pak::pkg_install("dplyr", dependencies = TRUE)
+          pak::pkg_install("readr", dependencies = TRUE)
+          pak::pkg_install("stringr", dependencies = TRUE)
+          pak::pkg_install("purrr", dependencies = TRUE)
+          pak::pkg_install("glue", dependencies = TRUE)
```
cpsievert, line 64: This way we can freely add/remove pkgs from the `vignettes/` folder

```suggestion
          pkgs <- unique(renv::dependencies("vignettes")$Package)
          lapply(pkgs, function(pkg) {
            if (! (pkg %in% rownames(installed.packages()))) {
              pak::pkg_install(pkg, upgrade = TRUE, dependencies = TRUE)
            }
          })
```

vignettes/available-sass-variables.Rmd
nstrayer: I agree. I think especially the readers of this article will understand the need to prefix variable names with some sort of identifier for sass.

vignettes/available-sass-variables.Rmd
nstrayer: Specifically the "BS4 only" note?

vignettes/available-theming-variables.Rmd
```diff
@@ -0,0 +1,249 @@
...
+```
+
+```{r variables_df}
+# BS3
+# Not using BS3 now because the comment templates are different so a whole new scraping pipeline is needed
+# sass_variables_loc <- "https://raw.githubusercontent.com/rstudio/bslib/master/inst/lib/bs-sass/assets/stylesheets/bootstrap/_variables.scss"
+
+# BS4
+sass_variables_loc <- rprojroot::find_package_root_file("inst/lib/bs/scss/_variables.scss")
+sass_variables_url <- "https://github.com/rstudio/bslib/blob/master/inst/lib/bs/scss/_variables.scss"
```
cpsievert, line 58:
```suggestion
sass_variables_url <- file.path(
  "https://github.com/rstudio/bslib/blob", 
   substr(system("git rev-parse HEAD", intern = TRUE), 1, 7),
  "inst/lib/bs/scss/_variables.scss"
)
```

vignettes/available-theming-variables.Rmd
```diff
@@ -0,0 +1,249 @@
...
+"
+
+variables_df %>%
+  rename(
+    Variable = variable,
+    Value = value,
+    Comment = comment,
+    `Find in source` = declaration
+  ) %>% 
+  DT::datatable(
```
cpsievert, line 202: I think we may want to consider adding `style = "bootstrap"`

vignettes/available-theming-variables.Rmd
```diff
@@ -0,0 +1,249 @@
...
+  ) %>%
+  collapse_multiline_comments() %>%
+  mutate(
+    below_is_empty_comment = lead(is_empty_comment),
+    above_is_empty_line = lag(line) == "",
+    below_is_empty_line = lead(line) == "",
+    is_section = (is_comment & below_is_empty_comment) | (is_comment & above_is_empty_line & below_is_empty_line),
+    variable = str_extract(line, "^\\$(.+)?(?=:)"),
+    section = ifelse(is_section, str_remove(line, "\\/\\/\\s*"), NA),
+    section_link = ifelse(is_section, link_to_pos_in_script(line_number), NA),
```
cpsievert, line 113: Sorry for not thinking of this earlier, but instead of rendering everything in one giant table, how about we generate a different table for each section and navigate them with tabset pills? This would make the different sections much easier to navigate. Something like:

```r
tabsetPanel(
  type = "pills",
  tabPanel("Color system", DT::datatableOutput("color_system")),
  ...
)
```

(I'm not sure how easy this would be without Shiny, though, and I'm sure this complicate to the hyperlinking as well since you'd need some sort of map between sections and variables)
 

  should tabs for each section to display a different table

vignettes/available-sass-variables.Rmd
cpsievert: Oh, yea, just the note. Also, referencing the R docs above, could point to https://rstudio.github.io/bslib/articles/theming.html#sass-variables-1 instead

cpsievert: Relatedly, we should probably link to this table towards the end of that section

cpsievert: Would also be good to replace existing hyperlinks in the docs to point to the table instead, for example https://github.com/rstudio/bslib/blame/master/README.md#L61

vignettes/available-theming-variables.Rmd
cpsievert: Somewhat relatedly, I think it'd prefer scrollable DT table(s) over pagination. Most everyone is not going to know what they're looking for, so the more we can "surface" the content without having to click through it, the better

vignettes/available-theming-variables.Rmd
nstrayer: image
My only hesitation with this is if you build the vignette outside of pkgdown bootstrap isn't available and it causes this ugliness

nstrayer: Nvm, fixed this with `htmltools::tagList(rmarkdown::html_dependency_bootstrap("default"))`

vignettes/available-theming-variables.Rmd
nstrayer: I think a difficulty with the tabs also would be navigating between dependencies. Certainly possible, but a bit more logic to look up what table the dependency lives in, then navigate to that tab, then navigate to the right location

nstrayer: As for scrolling DT, I agree. I was using tabs because I couldn't get the scrolling to work properly earlier. I'll take another crack at it today though.

nstrayer: I got the scrolling table to work and I think it's way nicer. The dependency finding is much more intuitive with animated scrolling

vignettes/available-theming-variables.Rmd
```diff
@@ -0,0 +1,254 @@
...
+  font-size: 0.85rem;
+}
+table.dataTable tr[role="row"] td:nth-child(4) {
+  text-align: center;
+}
+.dep-link {
+  color:steelblue; 
+  text-decoration:underline;
+  cursor: pointer;
+}
```
cpsievert, line 48:
```suggestion
}
.selected .dep-link {
  color: white;
}
```

vignettes/available-theming-variables.Rmd
```diff
@@ -0,0 +1,254 @@
...
+    is_comment = str_detect(line, "^\\/\\/"),
+    is_empty_comment = str_detect(line, "^\\/\\/\\s*$")
+  ) %>%
+  collapse_multiline_comments() %>%
+  mutate(
+    below_is_empty_comment = lead(is_empty_comment),
+    above_is_empty_line = lag(line) == "",
+    below_is_empty_line = lead(line) == "",
+    is_section = (is_comment & below_is_empty_comment) | (is_comment & above_is_empty_line & below_is_empty_line),
+    variable = str_extract(line, "^\\$(.+)?(?=:)"),
```
cpsievert, line 115: Looks like these variables weren't parsed correctly

Screen Shot 2021-01-26 at 1 33 24 PM

vignettes/available-theming-variables.Rmd
```diff
@@ -0,0 +1,254 @@
...
+```{r, include = FALSE}
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>",
+  echo = FALSE, 
+  message = FALSE
+)
+
+
+htmltools::tagList(rmarkdown::html_dependency_font_awesome(), rmarkdown::html_dependency_bootstrap("default"))
```
cpsievert, line 21: Oh wow, TIL pkgdown provides Bootstrap JS but not the CSS (only if you've specified a Bootswatch theme)

vignettes/available-theming-variables.Rmd
cpsievert: Yes, this is much better, thanks!

vignettes/available-theming-variables.Rmd
cpsievert: I'll handle this shortly

vignettes/available-theming-variables.Rmd
```diff
@@ -0,0 +1,258 @@
...
+# e.g.
+# // Section ID
+#
+# Or
+# // Section ID
+# //
+#
+# Sometimes the sections are just isolated floating comments
+#
+# // Section ID
```
cpsievert, line 77: This worries me a little bit, maybe we should patch the source file so that all the sections adhere to:

```
// Section title
//
// Section description
```

cpsievert: This way, we'll have some sort of guarantee that we're not creating sections that we're not aware of

vignettes/available-theming-variables.Rmd
```diff
@@ -0,0 +1,258 @@
...
+    variable = str_extract(line, "^\\$(.+)?(?=:)"),
+    value = str_extract(line, "(?<=:)(.+)?(?=;)") %>% str_remove(fixed("!default")) %>% str_trim(),
+    value_is_function = str_detect(value, fixed("()")) | str_detect(value, "\\($"),
+    # map-get function is used to reference a map built earlier so make sure not
+    # to link to it as it's not available in the table
+    value = str_replace_all(
+      value,
+      "(?<!map-get\\()(\\$[\\w|-]+)",
+      "<span class='dep-link'>\\1</span>"
+    ),
```
cpsievert, line 143: Hmm, how come I see this?

Screen Shot 2021-01-26 at 2 02 31 PM

vignettes/available-theming-variables.Rmd
```diff
@@ -0,0 +1,258 @@
...
+      TRUE ~ ""
+    ) %>% str_remove_all("\\/\\/") %>% str_trim(),
+    comment = ifelse(str_detect(comment, "stylelint-disable"), "", comment)
+  ) %>%
+  tidyr::fill(section, section_link) %>%
+  filter(!is.na(variable)) %>%
+  mutate(
+    variable = str_extract(line, "^\\$(.+)?(?=:)"),
+    value = str_extract(line, "(?<=:)(.+)?(?=;)") %>% str_remove(fixed("!default")) %>% str_trim(),
+    value_is_function = str_detect(value, fixed("()")) | str_detect(value, "\\($"),
```
cpsievert, line 136: I'm not sure what this is doing (In Sass, `()` creates a map, and there are no functions in the variables file?)

Is it intentional that you've decided not to include relevant variables like `$spacers` and `$sizes`?

vignettes/available-theming-variables.Rmd
cpsievert: Oh I get it now

vignettes/available-theming-variables.Rmd
```diff
@@ -0,0 +1,258 @@
...
+      table.row(row_index).scrollTo();
+    }
+  }
+}
+"
+
+variables_df %>%
+  rename(
+    Variable = variable,
+    Value = value,
```
cpsievert, line 206: I think we may want to consider having both a `Sass value` (i.e., the current `Value` column) and `CSS value` (which would be `bslib::bs_get_variables(bslib::bs_theme(), variable)`)

cpsievert: Somewhat relatedly, it'll a little unfortunate that you currently have to scroll far to the right to get the source links, perhaps it'd be better to column them in the `Variable` column (or at least closer to it?)

vignettes/available-theming-variables.Rmd
nstrayer: Oops. I put the lazy-eval flag outside of the variable capture group so it was capturing in between the dollar sign and the last "`:`" in the line. Fixed.

vignettes/available-theming-variables.Rmd
nstrayer: It leaves `map-get` in, it just makes sure not to make `$grid-breakpoints` a linked dependency because there is no entry in the table to link to.

vignettes/available-theming-variables.Rmd
nstrayer: Yeah. I just left out any of the variables that were functions because some of them have crazy long definitions and seem potentially confusing. For instance the definintion of `$grays`:
```scss
$grays: () !default;
$grays: map-merge(
  (
    "100": $gray-100,
    "200": $gray-200,
    "300": $gray-300,
    "400": $gray-400,
    "500": $gray-500,
    "600": $gray-600,
    "700": $gray-700,
    "800": $gray-800,
    "900": $gray-900
  ),
  $grays
);
```
That being said, I'm definitely not sure this is the correct way.

cpsievert: Ok, yea, I think OK with not including maps in the table, but we should probably mention them somewhere else

vignettes/available-theming-variables.Rmd
```diff
@@ -0,0 +1,258 @@
...
+  DT::datatable(
+    escape = FALSE,
+    style = "bootstrap",
+    rownames = FALSE,
+    selection = "none",
+    extensions = c("Select", "RowGroup", "Scroller"),
+    callback = DT::JS(glue("
+      [[js_helper_fns]]
+      var desired_var = location.hash.replace('#', '');
+      find_and_select_variable(table, desired_var, true);",
```
cpsievert, line 235: Can we also call this when the hash changes (so we can have hyperlinks within the article)? https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onhashchange

nstrayer: A good idea! Done in https://github.com/rstudio/bslib/pull/238/commits/602e2f26d9c9b5655808c005389edfa697bc4a87

nstrayer (commented):
Accidentally committed my .gitignore

vignettes/available-sass-variables.Rmd
cpsievert: I'll addressing these things throughout the day

vignettes/available-theming-variables.Rmd
nstrayer: image

Now the compiled css value can be seen on hover of the sass value using a tooltip. I think it's kinda slick.

vignettes/available-theming-variables.Rmd
nstrayer: Done here: https://github.com/rstudio/bslib/pull/238/commits/2256910d1a7cdd56f7aeb37d32f6a782b845fd72

PR #244, opened 2021-01-29
Wider sass variables table 

nstrayer:
Now the table is 90% width so users can actually see most or all of the columns. Has some trickery to detect when in a pkgdown environment and adjust to the weird layout structure.


cpsievert:
Closing in favor of https://github.com/rstudio/bslib/pull/245 (for cleaner git history)

For future reference, to avoid the cluttered history, it'd be better to `git rebase master` on this branch rather than `git merge master`

PR #248, opened 2021-02-02
made sure table doesn't just grow to huge on ultra-wide displays

nstrayer:
Before the sass variables table would always be 90% of the available viewport wide. This is good until you're working on a super-wide display and the table is out-of-control. Now we max its width at `1300px` which lets all columns be seen.


cpsievert:
Thanks!

Issue #282, opened 2021-02-24
Allow *-font-family-* variables to understand font helpers


cpsievert:
For example, `bs_theme("input-font-family" = font_google("Pacifico"))`

cpsievert:
Also, port these font importing helpers to sass, so packages like distill can use them (cc @apreshill)

PR #291, opened 2021-03-19
Generalized font importing

### LINKED ISSUE #256: simplify yaml front matter when used in Rmd to provide fonts
Currently, the way to provide some fonts is to use specific function when you want google fonts for example. 
This lead to this type of Rmd header
```yaml
   theme:
      base_font: !expr bslib::font_google("Prompt")
      code_font: !expr bslib::font_google("JetBrains Mono")
```

For now, we manage to avoid the use of `!expr` for basic feature and keep it for only advanced customisation (most of the time undocumented). Mainly because it is not easy to write, understand and it is a bit confusing with inline knitr code chunk. 

I would love to be able to write this instead
```yaml
   theme:
      base_font: 
        google: Prompt
      code_font: 
        google: JetBrains Mono
```
or 
```yaml
   theme:
     base_font: 
       google: 
         family: Prompt
         wght: "200..900"
```

Fonts is something that I believe users will want to modify easily and having a more YAML like syntax not relying on `!expr` would be nice. 

cc @yihui

cpsievert:
This PR is essentially a follow up to https://github.com/rstudio/sass/pull/74 (i.e., it reexports `sass::font_google()` et. al), and by consequence, `font_google()` et. al. now "just work" with any font-family Bootstrap Sass variable (closes https://github.com/rstudio/bslib/issues/282)

```r
bs_theme_preview(
  bs_theme("input-font-family" = font_google("Pacifico"))
)
```

Also, since `font_collection()` transforms `font_collection(google = "Pacifico")` into `font_collection(font_google("Pacifico"))` this PR also closes #256


R/bs-theme-update.R
```diff
@@ -200,29 +200,36 @@ bs3_accent_colors <- function(args) {
...
+    return(NULL)
+  }
+  # sass::font_collection handles quoting
+  if (sass::is_font_collection(x)) {
+    return(x)
+  }
+  # Unquote strings and provide to font_collection()
+  if (is.character(x)) {
+    families <- strsplit(x, ",\\s*")[[1]]
+    families <- gsub("'", "", gsub('"', "", families))
```
cpsievert, line 223: Trim white space too

R/imports.R
```diff
@@ -23,4 +25,66 @@ NULL
...
+#'   bs_theme_preview(theme)
+#' }
+NULL
+
+
+#' @importFrom sass font_link
+#' @export font_link
+#' @name font_link
+#' @rdname font_face
+NULL
```
cpsievert, line 77:
```suggestion
NULL
```
```suggestion
#' @importFrom sass font_link
#' @rdname font_face
#' @export
sass::font_link
```

R/bs-theme-update.R
cpsievert: Consider moving this logic to sass too, so that 

```r
sass(list(base_font = font_collection("foo, 'foo bar'"))
```

is supported

R/bs-theme-update.R
```diff
@@ -200,29 +200,36 @@ bs3_accent_colors <- function(args) {
...
+    return(do.call(sass::font_collection, as.list(families)))
+  }
+  if (is.list(x)) {
+    # TODO: have special awareness of named lists (e.g., font_base = list(google = list(family = "Pacifico")) => font_base = font_google("Pacifico")?)
+    do.call(sass::font_collection, lapply(x, prepare_font_arg))
```
cpsievert, line 228: Move this to `{sass}` as well

DESCRIPTION
```diff
@@ -55,13 +55,12 @@ Collate:
...
     'imports.R'
     'onLoad.R'
     'precompiled.R'
     'shiny-devmode.R'
     'version-default.R'
     'versions.R'
-    'web-font.R'
 URL: https://rstudio.github.io/bslib/, https://github.com/rstudio/bslib
 BugReports: https://github.com/rstudio/bslib/issues
+Remotes: rstudio/sass#74
```
cpsievert, line 66: Remove once sass PR is merged

R/bs-theme-layers.R
```diff
@@ -95,14 +95,27 @@ bs_add_variables <- function(theme, ..., .where = "defaults", .default_flag = id
...
+            "the ", nm, " variable (since it has it's own `default_flag`)."
+          )
+        }
+        return(val)
+      }
+      val <- paste(as_sass(val), collapse = "\n")
+      if (grepl("!default\\s*;*\\s*$", val)) {
+        val
+      } else {
+        paste(sub(";+$", "", val), "!default")
```
schloerke, line 115: Is the ending `;` always there? (If it comes from `sass`, then prolly yes.)

If not...

trim white space as well

```suggestion
        paste(sub("\\s*;*\\s*$", " ", val), "!default;")
```

Or 

also append the `;`

```suggestion
        paste(sub(";+$", "", val), "!default;")
```

schloerke (approved):

R/bs-theme-layers.R
cpsievert: Appending the `;` actually isn't necessary since the result will run through `sass::as_sass.list` again, which adds the trailing comma

(That being said, `sass::as_sass.list` should probably only add the `;` if it's necessary)

PR #295, opened 2021-03-30
Revert "no mo .rda file! (#294)"

cpsievert:
Reverts #294 since it increases the built tarball from about 4.1 Mb to 5.1 Mb


wch:
FWIW, I see built sizes of 4263141 bytes for the .R version, and 4236505 for the .rda version.


This is with the .R file:

```
$ cd bslib && git checkout aa77ed7 && cd .. && R CMD build bslib && ls -l bslib_0.2.4.9002.tar.gz
HEAD is now at aa77ed7  Add a font_scale argument to bs_theme() for a cross-version scalar multiplier of the base font size (#288)
....

-rw-r--r--  1 winston  staff  4263141 Mar 30 18:34 bslib_0.2.4.9002.tar.gz
```

And with the .rda file:

```
$ cd bslib && git checkout a193910 && cd .. && R CMD build bslib && ls -l bslib_0.2.4.9002.tar.gz
HEAD is now at a193910 Revert "no mo .rda file! (#294)" (#295)
....

-rw-r--r--  1 winston  staff  4236505 Mar 30 18:33 bslib_0.2.4.9002.tar.gz
```

PR #305, opened 2021-04-09
Fix typos

apreshill:
Hi @cpsievert -

I think these are corrections- if I'm wrong especially on the font variables please let me know! But I'm referencing these other sections: https://rstudio.github.io/bslib/articles/bslib.html#main-colors and https://rstudio.github.io/bslib/reference/bs_theme.html


cpsievert:
Thank you!

PR #306, opened 2021-04-15
Support .dropdown-item wrapped in <li> tags

cpsievert:
Ports https://github.com/twbs/bootstrap/pull/33649 into the v4.6.0 release. 

This'll allow shiny to use the same HTML structure for tabset dropdowns by requiring `bslib (>= 0.2.4.9003)`


wch (approved):
Looks good to me. What's the change in sysdata.rda?

cpsievert:
An update to the internal Google Fonts database (used for lookup inside the real-time themer when fonts are entered)

wch:
Opaque changes like that make me think we should reinstate #294. The package size increased by only 26kB in my testing.

PR #308, opened 2021-04-16
Allow for bootswatch='default' in bs_theme()

cpsievert:
Will be useful for https://github.com/rstudio/flexdashboard/pull/327, which wants to use Bootswatch cosmo by default

### DIFF
--- a/tests/testthat/test-theme-bootswatch.R
+++ b/tests/testthat/test-theme-bootswatch.R
@@ -22,7 +22,7 @@ test_that("Can retrieve version from theme object", {
     sass::sass(bs_theme(version = "4"))
   )
 
-  # Can use default as a wat to "explicitly don't use" a Bootswatch theme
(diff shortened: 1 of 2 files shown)


tests/testthat/test-theme-bootswatch.R
```diff
@@ -15,4 +15,15 @@ test_that("Can retrieve version from theme object", {
...
     theme_bootswatch(theme), "darkly"
   )
+
+  # Can use default to effectively remove bootswatch
+  expect_identical(
+    sass::sass(bs_theme_update(theme, bootswatch = "default")),
+    sass::sass(bs_theme(version = "4"))
+  )
+
+  # Can use default as a wat to "explicitly don't use" a Bootswatch theme
```
cpsievert, line 25:
```suggestion
  # Can use default as a way to "explicitly don't use" a Bootswatch theme
```

PR #334, opened 2021-06-16
Fix condition expectations in prevision of testthat breaking change

lionel-:
We are planning a breaking change to the condition expectations in testthat release (see NEWS item in https://github.com/r-lib/testthat/pull/1401). This PR implements a preventive fix that works with and without the breaking change.

There is no hurry to get this fix on CRAN since we are skipping one released version of testthat before going through with the change.


cpsievert:
Thanks!

PR #339, opened 2021-06-23
Throw version=5 only when dependencies are requested and shiny is loaded

### LINKED ISSUE #338: warning thrown about version  5 and shiny requirement even when not asked
While working on `bookdown::bs4_book()` which is using version 4 of bootstrap, I noticed I got a warning thrown

````
Warning message:
`bs_theme(version = 5)` is designed to work with shiny v1.6.0.9001 or higher 
````

This comes from `precompiled_css()` which is looping all version even when not asked by the user.

``` r
x <- bslib:::precompiled_css()
#> Warning: `bs_theme(version = 5)` is designed to work with shiny v1.6.0.9001 or
#> higher
```

This will be fixed by a shiny cran release probably, but it is a bit disturbing as we are looking at where it could go wrong. 

maybe `suppressWarnings()` could be used, or a special handling of this specific warning ? 

Anyway, just wanted to share this as I looked into it thinking there was an issue somewhere in **bookdown**.

cpsievert:
Closes #338

### DIFF
--- a/R/bs-dependencies.R
+++ b/R/bs-dependencies.R
@@ -65,7 +65,7 @@ bs_theme_dependencies <- function(
   theme <- as_bs_theme(theme)
   version <- theme_version(theme)
 
-  if (isTRUE(version == 5)) {
+  if (isTRUE(version >= 5)) {
     shiny_version <- "1.6.0.9001"
     msg <- sprintf("`bs_theme(version = 5)` is designed to work with shiny %s or higher", shiny_version)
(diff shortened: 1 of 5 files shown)


R/bs-dependencies.R
```diff
@@ -65,6 +65,15 @@ bs_theme_dependencies <- function(
   theme <- as_bs_theme(theme)
   version <- theme_version(theme)
 
+  if (isTRUE(version == 5)) {
```
cpsievert, line 68:
```suggestion
  if (isTRUE(version >= 5)) {
```

PR #405, opened 2021-11-04
Various speed improvements

### LINKED ISSUE #369: Performance issue with selectize
Hi, when building a large app with many inputs, I noticed a performance issue.

```
library(shiny)
library(bslib)

ui <- fluidPage(
  theme = bs_theme(bootswatch = "slate"),
  do.call("tagList", lapply(1:200, function(x) {
    selectInput(paste0("id", x), "Another input", choices = LETTERS, selectize = TRUE)
  }))
)
server <- function(input, output, session) {}
shinyApp(ui, server)
```

Profiling the app with

```
profvis::profvis(runApp('app.R'))
```

shows that `bslib::bs_dependency()` is called many times and causes the slowdown. Total time: 6580ms on my computer.

The time goes down to 1320ms if `selectize = FALSE` or if bslib is not used.

Related question: https://stackoverflow.com/q/67726350 (I didn't ask that question but I took its reprex)

`sessionInfo()`
```
R version 4.1.0 (2021-05-18)
Platform: x86_64-apple-darwin17.0 (64-bit)
Running under: macOS Big Sur 11.5.2

Matrix products: default
LAPACK: /Library/Frameworks/R.framework/Versions/4.1/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices datasets  utils     methods   base     

other attached packages:
[1] bslib_0.3.0 shiny_1.6.0

loaded via a namespace (and not attached):
 [1] Rcpp_1.0.7        magrittr_2.0.1    xtable_1.8-4      R6_2.5.0         
 [5] rlang_0.4.11      fastmap_1.1.0     fansi_0.5.0       stringr_1.4.0    
 [9] tools_4.1.0       sessioninfo_1.1.1 utf8_1.2.2        cli_3.0.1        
[13] withr_2.4.2       jquerylib_0.1.4   htmltools_0.5.2   ellipsis_0.3.2   
[17] yaml_2.2.1        digest_0.6.27     tibble_3.1.3      lifecycle_1.0.0  
[21] crayon_1.4.1      later_1.2.0       vctrs_0.3.8       profvis_0.3.7    
[25] sass_0.4.0        htmlwidgets_1.5.3 promises_1.2.0.1  fs_1.5.0         
[29] cachem_1.0.5      mime_0.11         stringi_1.7.3     pillar_1.6.2     
[33] compiler_4.1.0    jsonlite_1.7.2    httpuv_1.6.1      renv_0.14.0      
[37] pkgconfig_2.0.3  
```

cpsievert:
Closes #369


cpsievert:
@wch the CI should pass once we've merged https://github.com/wch/staticimports/pull/11

R/bs-dependencies.R
```diff
@@ -307,24 +313,37 @@ bs_dependency <- function(input = list(), theme, name, version,
...
+bs_dependency_defer <- function(func, memoise = TRUE) {
+  # func() most likely calls stuff like sass_file() and bs_dependency() ->
+  # sass_partial() -> sass() (e.g., see example section above) Even though
+  # sass() calls can be cached, there is still considerable overhead involved
+  # in computing the cache key itself because it involves file-level operations
+  # (e.g. getting the mtimes of file attachments).
+  # So, to help with performance (in the case where we repeatedly call the same
+  # func with the same theme), memoise func with a short-lived cache
+  if (memoise) {
+    mfunc <- memoise::memoise(func, cache = .dependency_cache)
```
wch, line 325: This will create a new memoized function each time `bs_dependency_defer()` is called -- so, for example, every time `selectizeInput()` is called, it will get a separate memoized function. They _do_ share the caching object, so each copy will benefit from caching, but it's extra work to do the memoization every time this function is called.

I think there may be a way to just memoize the function once, but exactly how isn't clear to me just yet...

cpsievert: > it's extra work to do the memoization every time this function is called

FWIW, the overhead is 10ms in the case of example provided in #369:

Screen Shot 2021-11-17 at 10 49 09 AM


It seems like getting this right is gonna make this already complicated thing much more complicated, so it doesn't seem worth it to me

wch:
```suggestion
    # The memoized function is created each time bs_dependency_defer is called,
    # and then it is used once. This is not how memoized functions are normally
    # used, but in this case it works because the caching object is re-used, and
    # it still provides very significant improvement in performance.
    mfunc <- memoise::memoise(func, cache = .dependency_cache)
```

wch (approved):

Issue #360, opened 2021-08-13
Improve docs describing how to add custom CSS


hedsnz:
Extending the bslib themes with some custom CSS seems like a common use case which I thought would be documented early. It took me quite a long time to find out how to actually do this; ultimately I found it [here](https://rstudio.github.io/bslib/articles/utility-classes.html), with `bs_add_rules`,

```
bs_theme() %>% 
    bs_add_rules("#my-nav { @extend .justify-content-center }")
```

(I was confused why I couldn't add custom CSS in the way I'm accustomed to with Shiny, e.g., through `tags$head(tags$style(HTML("...")))` or `includeCSS`, etc.)

I think it would be helpful to show how to add custom CSS earlier on in the docs, ideally in the Get Started vignette. Maybe [here](https://rstudio.github.io/bslib/articles/bslib.html#add-rules), alongside the `@import` example.

This may just be a reflection of my reading comprehension, so feel free to close this issue if you think it's not a big deal. I'd be happy to submit a PR for this. This could either add to the `@extend` example above, or a simple new example, e.g.,

```
navbarPage(
    theme = bs_theme() %>%
        bs_add_rules("
            .navbar-header {
                position: absolute;
            }
            ul.nav.navbar-nav { 
                display: flex;
                justify-content: center;
            }
        "),
    title = "Centered navbar",
    tabPanel("One"),
    tabPanel("Two")
)
```

cpsievert:
In general I think it's a good idea to provide another example of `bs_add_rules()`, but I'd like to avoid this example now that we have `bslib::nav_spacer()` (i.e., you don't need extra CSS like this to center/right align nav items anymore)

cpsievert:
The new gettting started with theming article does a better job of this https://rstudio.github.io/bslib/articles/theming.html#add-rules

PR #411, opened 2022-02-08
docs: example of include CSS using bs_add_rules

hedsnz:
Addresses https://github.com/rstudio/bslib/issues/360.
Something like this?


cpsievert:
See https://github.com/rstudio/bslib/issues/360#issuecomment-1575030491

PR #433, opened 2022-06-21
Add card() API

cpsievert:


R/card.R
```diff
@@ -0,0 +1,298 @@
...
+}
+
+#' @rdname card_body
+#' @export
+card_title <- function(..., class = NULL) {
+  res <- htmltools::tags$h5(class = "card-header", class = class,
+    ...
+  )
+  as.card_item(res)
+}
```
cpsievert, line 237: Did you intend to have 3 versions of essentially the same thing? I think I would be in favor of just having  (especially since you can easily control the size with the `.display-*` utility classes)

```suggestion
#' @rdname card_body
#' @export
card_header <- function(..., class = NULL, container = htmltools::div) {
  res <- container(class = "card-header", class = class,
    ...
  )
  as.card_item(res)
}
```

R/card.R
```diff
@@ -0,0 +1,298 @@
...
+
+#' @rdname card_body
+#' @export
+card_footer <- function(..., class = NULL) {
+  res <- htmltools::tags$div(class = "card-footer mt-auto", class = class,
```
cpsievert, line 242:
```suggestion
  res <- htmltools::div(class = "card-footer mt-auto", class = class,
```

R/utils-tags.R
```diff
@@ -2,7 +2,7 @@ tag_require <- function(tag, version = 4, caller = "") {
   tagAddRenderHook(
     tag, replace = FALSE,
     func = function(x) {
-      current_version <- theme_version(bs_current_theme()) %||% 3
+      current_version <- theme_version(bs_current_theme()) %||% version_default()
```
cpsievert, line 5: Why was this needed? Seems like this could give wrong results if this gets called and `bs_theme()` isn't being used for the Bootstrap dependency

R/card.R
```diff
@@ -0,0 +1,298 @@
...
+
+  # TODO: card-img-* needs to go on the <img> itself, not the containing <div>
+  plot_div <-
+    tagAppendAttributes(plot_div,
+      style = css(
+        flex = if (stretch) "1 1",
+        `-webkit-flex` = if (stretch) "1 1",
+        # May be NULL
+        `flex-basis` = validateCssUnit(height),
+        `-webkit-flex-basis` = validateCssUnit(height),
```
cpsievert, line 292: What made you add the `-webkit-*` prefix? Is it only because `bs_theme()` adds them? Since those are added only for phantomjs (`{shinytest}`) support, I'd actually prefer that we don't add them here (more motivation to migrate to `{shinytest2}`)

R/card.R
```diff
@@ -0,0 +1,298 @@
...
+  ...
+) {
+  plot_div <- plotOutput(outputId, height = NULL, click = click, dblclick = dblclick, hover = hover,
+    brush = brush)
+
+  # TODO: card-img-* needs to go on the <img> itself, not the containing <div>
+  plot_div <-
+    tagAppendAttributes(plot_div,
+      style = css(
+        flex = if (stretch) "1 1",
```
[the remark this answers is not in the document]
cpsievert: Also, in general, I don't like the two-value syntax since the meaning of the 2nd number changes depending on the value https://developer.mozilla.org/en-US/docs/Web/CSS/flex#syntax

R/card.R
cpsievert, line 292: I find the two-value syntax confusing since the interpretation of the 2nd number changes depending on the value's unit https://developer.mozilla.org/en-US/docs/Web/CSS/flex#syntax

What made you add the `-webkit-*` prefix? Is it only because `bs_theme()` adds them? Since those are added only for phantomjs (`{shinytest}`) support, I'd actually prefer that we don't add them here (more motivation to migrate to `{shinytest2}`)


```suggestion
        flex_grow = if (stretch) "1",
        flex_shrink = if (stretch) "1",
        # May be NULL
        flex_basis = validateCssUnit(height),
```

R/card.R
```diff
@@ -0,0 +1,298 @@
...
+# card_list <- function(...) {
+#   res <- htmltools::tags$ul(class = "list-group list-group-flush", ...)
+#   as.card_item(res)
+# }
+#
+# #' @export
+# card_list_item <- function(...) {
+#   htmltools::tags$li(class = "list-group-item", ...)
+# }
+
```
cpsievert, line 269: Move this to the bottom

R/card.R
```diff
@@ -0,0 +1,298 @@
...
+#' @export
+card_body_scroll <- function(..., class = NULL, height = NULL, padding = c("x", "y")) {
+  res <- htmltools::tags$div(class = "card-body card-body-scroll",
+    class = if (!"x" %in% padding) "px-0",
+    class = if (!"y" %in% padding) "py-0",
+    class = class,
+    style = css(
+      # May be NULL
+      flex_basis = validateCssUnit(height),
+      `-webkit-flex-basis` = validateCssUnit(height)
```
cpsievert, line 205:
```suggestion
      flex_basis = validateCssUnit(height)
```

R/card.R
```diff
@@ -0,0 +1,298 @@
...
+card_body <- function(..., class = NULL, padding = c("x", "y"), stretch = FALSE) {
+  res <- htmltools::tags$div(class = "card-body",
+    class = if (!"x" %in% padding) "px-0",
+    class = if (!"y" %in% padding) "py-0",
+    class = class,
+    ...,
+    style = if (!stretch) css(
+      flex = "0",
+      `_webkit_flex` = "0"
+    ),
```
cpsievert, line 191:
```suggestion
    style = if (!stretch) css(flex = "0"),
```

R/card.R
```diff
@@ -0,0 +1,298 @@
...
+#' @param class Additional CSS classes to include on the card div.
+#' @param padding Whether vertical and/or horizontal padding should be included.
+#'   Use `NULL` for neither, `"x"` for horizontal only, `"y"` for vertical only,
+#'   or `c("x", "y")` for both (the default).
+#' @param stretch Set to `TRUE` if this `card_body` is eager to use any extra
+#'   vertical space is available in the card.
+#'
+#' @export
+card_body <- function(..., class = NULL, padding = c("x", "y"), stretch = FALSE) {
+  res <- htmltools::tags$div(class = "card-body",
```
cpsievert, line 183:
```suggestion
  res <- htmltools::div(class = "card-body",
```

R/card.R
```diff
@@ -0,0 +1,298 @@
...
+#'   grid will have the same height. If `"row"`, then every card in _each_ row
+#'   of the grid will have the same height, but heights may vary between rows.
+#' @param fixed_width When `card_width` is a CSS length (as opposed to a
+#'   fraction), the default behavior is to let cards be that wide _or larger_,
+#'   so that they fill all of the horizontal space that's available. Set this
+#'   argument to `TRUE` to ensure that cards are exactly the specified width,
+#'   even if this means using less than the total width of the card grid
+#'   container.
+#'
+#' @export
```
cpsievert, line 30: We should also include `@example` and/or link to the article

```suggestion
#' @export
#' @family cards
```

R/card.R
```diff
@@ -0,0 +1,298 @@
...
+#'   element. Unnamed arguments become card items, and can be any valid child of
+#'   an [htmltools tag][htmltools::tags].
+#' @param class Additional CSS classes to include on the card div.
+#' @param padding Whether vertical and/or horizontal padding should be included.
+#'   Use `NULL` for neither, `"x"` for horizontal only, `"y"` for vertical only,
+#'   or `c("x", "y")` for both (the default).
+#' @param stretch Set to `TRUE` if this `card_body` is eager to use any extra
+#'   vertical space is available in the card.
+#'
+#' @export
```
cpsievert, line 181: We should also include `@example` and/or link to the article

```suggestion
#' @export
#' @family cards
```

R/card.R
```diff
@@ -0,0 +1,298 @@
...
+
+  # TODO: Allow cards not to stretch
+
+  heights_equal <- match.arg(heights_equal)
+
+  args <- rlang::list2(...)
+  argnames <- names(args)
+  if (is.null(argnames) && length(args) > 0) {
+    argnames <- rep_len("", length(args))
+  }
```
cpsievert, line 42:
```suggestion
  args <- rlang::list2(...)
  argnames <- rlang::names2(args)
```

R/card.R
```diff
@@ -0,0 +1,298 @@
...
+      flex = "0",
+      `_webkit_flex` = "0"
+    ),
+  )
+  as.card_item(res)
+}
+
+#' @export
+card_body_scroll <- function(..., class = NULL, height = NULL, padding = c("x", "y")) {
+  res <- htmltools::tags$div(class = "card-body card-body-scroll",
```
cpsievert, line 198:
```suggestion
  res <- htmltools::div(class = "card-body card-body-scroll",
```

R/card.R
```diff
@@ -0,0 +1,298 @@
...
+
+  div(class = "bslib-card-grid", class = class,
+    style = css(
+      grid_template_columns = colspec,
+      grid_auto_rows = if (heights_equal == "all") "1fr",
+      gap = validateCssUnit(gap)
+    ),
+    !!!attribs,
+    children
+  )
```
cpsievert, line 75:
```suggestion
  tag <- div(class = "bslib-card-grid", class = class,
    style = css(
      grid_template_columns = colspec,
      grid_auto_rows = if (heights_equal == "all") "1fr",
      gap = validateCssUnit(gap)
    ),
    !!!attribs,
    children
  )
  
  as_fragment(
    tag_require(tag, version = 4, caller = "card_grid()")
  )
```

R/card.R
```diff
@@ -0,0 +1,298 @@
...
+#'   `$spacer` Sass variable is used.
+#' @param heights_equal If `"all"` (the default), every card in every row of the
+#'   grid will have the same height. If `"row"`, then every card in _each_ row
+#'   of the grid will have the same height, but heights may vary between rows.
+#' @param fixed_width When `card_width` is a CSS length (as opposed to a
+#'   fraction), the default behavior is to let cards be that wide _or larger_,
+#'   so that they fill all of the horizontal space that's available. Set this
+#'   argument to `TRUE` to ensure that cards are exactly the specified width,
+#'   even if this means using less than the total width of the card grid
+#'   container.
```
cpsievert, line 28:
```suggestion
#' @param fixed_width Whether or not to interpret the `card_width` as a minimum
#'   (`fixed_width=FALSE`) or fixed (`fixed_width=TRUE`) width when it is a CSS
#'   length unit.
```

R/card.R
```diff
@@ -0,0 +1,298 @@
...
+#'
+#' @export
+card_grid <- function(..., class = NULL, card_width = 1/4, gap = NULL,
```
cpsievert, line 31: As much as I do like the idea of have a formal `class` argument, it is kind of a bummer that, this way, you can't provide multiple `class` arguments

R/card.R
jcheng5: I think I intended to have two (one generic `card-header` and one that takes a string and puts it in a heading) but not two. Certainly `card_heading` and `card_title` seem redundant. I'm open to whatever though.

cpsievert:
For sake of brevity, I'm closing this in favor of #452 (which incorporates but also extends some of the ideas here)

PR #475, opened 2023-01-03
Add accordion() API

### LINKED ISSUE #464: `accordion()` + `accordion_item()` missing from bslib
Hi Carson,

Have been using the versions demo'd at RStudio Conference for the past few months and noticed after updating that the `accordion()` and `accordion_item()` functions are now gone. Those two, along with everything else in [the demo](https://github.com/rstudio/bslib/blob/df1ccbc08d6f356daba947499df3b1c8127e2568/inst/examples/demo/app.R), were working really well  until I just updated today. They were really great! Have they been removed from the new version of bslib?

Thanks!

cpsievert:
Closes #464

### TODO (before merge)

- [x] Merge #396 
   * That PR attempts to address `margin-top` header issues in general, but we could also just add `.accordion-header { margin-top: 0 }`

- [x] Have @gregswinehart provide some different default styling?

### TODO (after merge)

- [ ] Add to docs of `insertTab()` about `target = NULL` behavior
- [ ] Get automated tests merged https://github.com/rstudio/shinycoreci/tree/bslib-accordion-sidebar


srcts/build/index.ts
```diff
@@ -0,0 +1,21 @@
+// This build script must be executed from the root repo directory via
+// ```
+// yarn build
+// ```
+
+import { build } from "./_build";
+import type { BuildOptions } from "esbuild";
+
+const opts: BuildOptions = {
+  target: ["es6"],
```
cpsievert, line 10: Add a comment about this decision

inst/components/accordion.scss
```diff
@@ -0,0 +1,14 @@
+$accordion-box-shadow-active: none !default;
```
jcheng5, line 1: Don't declarations like this belong in the `defaults` part of a `sass_layer`?

cpsievert: Good point, thanks!

jcheng5 (approved):

PR #479, opened 2023-01-24
Add `sidebar()`, `layout_sidebar()`, `card_sidebar()`, and `container()`

cpsievert:


inst/components/fill.scss
```diff
@@ -0,0 +1,18 @@
...
+  >.active.html-fill-container {
+    display: flex;
+  }
+
+  // Another workaround for pkgdown adding extra padding we didn't ask for
+  // https://github.com/r-lib/pkgdown/blob/956f07/inst/BS5/assets/pkgdown.scss#L335-L337
+  &.html-fill-container {
+    padding: 0;
+  }
+}
```
cpsievert, line 18: The only change here is that these rules are no longer scoped under `.bslib-card` (we still want `.tab-content>.active.html-fill-container` to apply to other tabs, notably for `page_navbar(fill = TRUE)`)

R/accordion.R
```diff
@@ -151,7 +151,7 @@ accordion_panel <- function(title, ..., value = title, icon = NULL) {
     class = "accordion-item",
     "data-value" = value,
     # TODO: can we provide a way to put more stuff in the header? Like maybe some right-aligned controls?
-    h2(class = "accordion-header", btn),
+    span(class = "accordion-header h2", btn),
```
jcheng5, line 154: Is this bad for accessibility?

R/navs.R
```diff
@@ -87,16 +88,44 @@ collect_nav_items <- function(..., wrapper) {
...
+  as.card_item(content)
+}
 
-  tabs <- bindFillRole(tabs, .cssSelector = ".tab-content", container = TRUE, item = TRUE)
-  tabs <- bindFillRole(tabs, .cssSelector = ".tab-content > *", container = TRUE, item = TRUE)
 
-  content <- tagQuery(tabs)$find(".tab-content")$selectedTags()
+# Given a .tab-content container, mark each relevant .tab-pane as a
+# fill container/item.
+fill_tab_content <- function(content, fill = FALSE, navbar_margin = FALSE) {
```
jcheng5, line 105: Left off here

jcheng5:
@cpsievert and I discussed an alternate way of expressing "this is(n't) a fill container" and "this is(n't) a fill item" (since `fill=TRUE` feels pretty ambiguous). Basically, using the word "fill" less, and having containers and items use different terms for their respective subtypes.

https://gist.github.com/jcheng5/d6066b5a07302b1fe9d354267eb7054d

inst/components/sidebar.scss
```diff
@@ -0,0 +1,214 @@
...
+      border: var(--bslib-collapse-toggle-border);
+      border-radius: 0 0 var(--bs-border-radius) var(--bs-border-radius);
+      padding: 0 4px;
+    }
+
+    &.sidebar-collapsed {
+      --bslib-collapse-toggle-transform: 0deg;
+      --bslib-collapse-toggle-right-transform: 0deg;
+
+      > .sidebar {
```
cpsievert, line 198: Can we transition height here?

srcts/src/components/sidebar.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  $container.addClass("transitioning");
+
+  $container.toggleClass(COLLAPSE_CLASS);
+  $side.trigger("toggleCollapse.sidebarInputBinding");
+});
+
+$(document).on("transitionend", ".bslib-sidebar-layout", (e) => {
+  const $el = $(e.target);
+  if ($el.hasClass("collapse-toggle")) {
```
cpsievert, line 73: Add a comment about why we're only doing this for transitions on collapse-toggle

cpsievert: Look into changing `".bslib-sidebar-layout"` to ".bslib-sidebar-layout .collapse-toggle"

inst/components/sidebar.scss
```diff
@@ -0,0 +1,214 @@
+$bslib-sidebar-padding: $spacer * 1.5 !default;
+$bslib-sidebar-icon-size: $spacer !default;
+
+.bslib-sidebar-layout {
+  --bslib-sidebar-bg: var(--bs-card-bg);
+  --bslib-sidebar-main-bg: var(--bs-card-bg);
```
cpsievert, line 6: Should these just be undefined by default?

cpsievert: If we respect bg, we also need to respect fg

jcheng5 (approved):

R/accordion.R
cpsievert: I didn't notice any difference in behavior in VoiceOver, so I think this is fine (and I'll add a comment about why we're not using `<h2>`)

PR #481, opened 2023-02-06
Optionally skip test file requiring shiny

MichaelChirico:


cpsievert:
Thanks, I think I'd prefer the skip to be within the `test_that()` context (since we decide to add more `page_()` that don't depend on Shiny).

PR #568, opened 2023-05-08
Use more verbose naming for fill helper functions

cpsievert:
As proposed by @schloerke. Follow up to #536


gadenbuie (approved):
This looks like a solid improvement to me!

The only one I'm on the fence about is `remove_all_fill()`, but I can't think of anything better.

Issue #569, opened 2023-05-09
Simplify dashboard-like layouts


cpsievert:
Right now, you'd need something like this for a (single-page) dashboard-like filling layout:

```r
page_fillable(
    padding = 0, gap = 0,
    h2("My dashboard", class = "bg-dark p-2 mb-0"),
    layout_sidebar(
        sidebar = sidebar("side"),
        fillable = TRUE,
        border = FALSE,
        border_radius = FALSE,
        plotly::plot_ly()
    )
)
```

We could do more to make smarter defaults for this special situation (i.e., a single `layout_sidebar()` appearing as a direct child of `page_fillable()`), but even if we do, that may not be good enough for a basic getting started tutorial. Instead of (or addition to) smarter defaults, we may want an higher-order abstraction like:

```r
page_sidebar(
    title = "My dashboard",
    sidebar = sidebar("side"),
    fillable = FALSE, # must opt-out of filling layout
    plotly::plot_ly()
)
```

This `page_sidebar()` API is pretty similar to how we currently recommend you create multi-page sidebar layouts. It might make sense to align these as much as possible (i.e., default to `fillable = TRUE` for both? Use `title` instead of `header` for both?)

```r
page_navbar(
    title = "My dashboard",
    sidebar = sidebar("side"),
    fillable = FALSE, # must opt-out of filling layout?
    nav_panel("Plot", plotly::plot_ly())
)
```

Also, once we have navigation in the sidebar (see #585), maybe an API like this could also work (we'd throw if you tried to supply `sidebar` or non-nav items as children and use `navset_sidebar()` under-the-hood):

```r
page_sidebar(
    title = "My dashboard",
    nav_panel("A", "a"),
    nav_panel("B", "b")
)
```

And, in the rare case you wanted both header and sidebar navigation, you'd have to start with `page_navbar()` (this should just fall out of #585):

```r
page_navbar(
  title = "My dashboard",
  nav_panel(
    "Page A", 
    navset_sidebar(
      nav_panel("Topic A", "Topic A"),
      nav_panel("Topic B", "Topic B")
    )
  ),
  nav_panel(
    "Page B", 
    navset_sidebar(
      nav_panel("Topic C", "Topic C"),
      nav_panel("Topic D", "Topic D")
    )
  )
)
```

cpsievert:
The core idea behind this is done now

PR #588, opened 2023-05-19
Add `page_sidebar()`

cpsievert:
Partially addresses #569 

## TODO

- [x] Tests 
    * https://github.com/rstudio/shinycoreci/pull/184 provides some coverage


R/page.R
```diff
@@ -73,6 +73,66 @@ validateCssPadding <- function(padding = NULL) {
...
+#'
+#' server <- function(input, output) {
+#'   output$p <- renderPlot({
+#'     ggplot(mtcars) + geom_histogram(aes(!!input$var))
+#'   })
+#' }
+#'
+#' shinyApp(ui, server)
+#'
+page_sidebar <- function(..., sidebar = sidebar(), title = NULL, fillable = TRUE, fill_mobile = FALSE, theme = bs_theme(), window_title = NULL, lang = NULL) {
```
cpsievert, line 112:
```suggestion
page_sidebar <- function(..., sidebar = sidebar(), title = NULL, fillable = TRUE, fill_mobile = FALSE, theme = bs_theme(), window_title = NA, lang = NULL) {
```

R/page.R
```diff
@@ -73,6 +73,66 @@ validateCssPadding <- function(padding = NULL) {
...
+
+  if (rlang::is_bare_character(title) || rlang::is_bare_numeric(title)) {
+    title <- h2(title, class = "bslib-page-title")
+  }
+
+  page_fillable(
+    padding = 0,
+    gap = 0,
+    title = get_window_title(title, window_title),
```
cpsievert, line 121: Pull this out from #589

inst/components/fill.scss
```diff
@@ -22,8 +22,18 @@
...
+
+.bslib-page-title {
+  background-color: $bslib-page-title-bg;
+  color: $bslib-page-title-color;
+  font-size: $h4-font-size;
+  font-weight: 300;
+  padding: $spacer * 0.5;
+  padding-left: $bslib-sidebar-padding;
+  margin-bottom: 0;
 }
```
cpsievert, line 37: Garrick to double-check this looks good / is sensible.

R/page.R
```diff
@@ -73,6 +71,73 @@ validateCssPadding <- function(padding = NULL) {
...
+#'     ggplot(mtcars) + geom_histogram(aes(!!input$var))
+#'   })
+#' }
+#'
+#' shinyApp(ui, server)
+#'
+page_sidebar <- function(..., sidebar = sidebar(), title = NULL, fillable = TRUE, fill_mobile = FALSE, theme = bs_theme(), window_title = NA, lang = NULL) {
+
+  if (rlang::is_bare_character(title) || rlang::is_bare_numeric(title)) {
+    title <- h2(title, class = "bslib-page-title")
```
gadenbuie, line 120: Given its role on the page, I think this should be an `<h1>`

```suggestion
    title <- h1(title, class = "bslib-page-title")
```

inst/components/fill.scss
```diff
@@ -22,8 +22,18 @@
...
+}
+
+.bslib-page-title {
+  background-color: $bslib-page-title-bg;
+  color: $bslib-page-title-color;
+  font-size: $h4-font-size;
+  font-weight: 300;
+  padding: $spacer * 0.5;
+  padding-left: $bslib-sidebar-padding;
+  margin-bottom: 0;
```
gadenbuie, line 36: Overall this is great! Given that both Bootstrap provides [three navbar color schemes](https://getbootstrap.com/docs/5.3/components/navbar/#color-schemes) and that Bootswatch highlights these in their theme previews, I think we might want to think about leaving some room for building on those classes in the future.

Maybe we could [take inspiration from quarto](https://quarto.org/docs/websites/website-navigation.html#top-navigation), who IIUC add some [sass variables](https://github.com/quarto-dev/quarto-cli/blob/28457e739a6fd1b3b1ae1bebe736aa9a487e1e84/src/resources/formats/html/bootstrap/_bootstrap-variables.scss#L41-L56) around navbar color and hook off of that. Maybe we'd hook into a sass variable that could be set easily via `bs_theme()`?

I think we're on a reasonable path that would allow for this kind of expansion, but I also think we can reasonably expect users to ask for something like this soon.

inst/components/fill.scss
```diff
@@ -22,8 +22,18 @@
...
+  padding: $spacer;
+  gap: $spacer;
+}
+
+.bslib-page-title {
+  background-color: $bslib-page-title-bg;
+  color: $bslib-page-title-color;
+  font-size: $h4-font-size;
+  font-weight: 300;
+  padding: $spacer * 0.5;
```
gadenbuie, line 34: I like this with more spacing and I think using `$spacer` gets closer to the `page_navbar()` sizing.

```suggestion
  padding: $spacer;
```


image

inst/components/fill.scss
cpsievert: > Maybe we could [take inspiration from quarto](https://quarto.org/docs/websites/website-navigation.html#top-navigation), who IIUC add some [sass variables](https://github.com/quarto-dev/quarto-cli/blob/28457e739a6fd1b3b1ae1bebe736aa9a487e1e84/src/resources/formats/html/bootstrap/_bootstrap-variables.scss#L41-L56) around navbar color and hook off of that. Maybe we'd hook into a sass variable that could be set easily via bs_theme()?

I'm pretty sure Quarto took inspiration from bslib :) We also have a [`$navbar-bg`](https://rstudio.github.io/bslib/articles/bs5-variables.html#navbar-bg), but I'm not sure I want to tie that variable to `page_sidebar()`. At least, for right now, folks who want to theme it can do `bs_theme("bslib-page-title-bg" = "black")`

cpsievert: Good point, Quarto actually got the `$navbar-bg` from bslib :)

PR #591, opened 2023-05-23
Use named argument when supplying `sidebar()` to `layout_sidebar()`

cpsievert:
After #588, the `sidebar` argument in `layout_sidebar()` must be explicitly named. I forgot to update the current uses of it accordingly


gadenbuie (approved):
I think seeing these changes validates that it's an overall good change!

PR #633, opened 2023-06-14
Update link to fontpair.co in theming article

asadow:
Link to [fontpair](fontpair.co) was broken. Fixed now, but maybe best to use the root as the site's Pairings tab is easily found.


gadenbuie (approved):
Thanks @asadow!

PR #677, opened 2023-07-14
Respect `bslib.precompiled` option when considering whether compile Sass for components

cpsievert:
Follow up to #664.


gadenbuie (approved):
🚀

PR #721, opened 2023-08-02
Prevent tooltip/popover contents from being pulled out of their custom element

### LINKED ISSUE #720: `tooltip()`/`popover()` doesn't work inside a `<p>` tag
Here's a minimal reprex:

```r
page_fixed(p(tooltip("A tooltip", "Message")))
```

The source of the issue is that browsers will actually render something like this:

```html
<p> A paragraph <div> A div </div> </p>
```

As

```html
<p> A paragraph </p>
<div> A div </div>
```

For (I assume) similar reasons, if you have a `<custom-element>` inside of a `<p>` tag, it appears _any_ HTML children of the `<custom-element>` will get similar treatment (even if all of them are `display:inline`). For example, this source HTML

```html
<p> A paragraph <custom-element> <span> A span </span> </custom-element> </p>
```

gets rendered as:

```html
<p> A paragraph <custom-element></custom-element> </p>
<span> A span </span>
```

This is quote unfortunate for the current implementation of `tooltip()`/`popover()`, which reaches into its children to get the relevant trigger and content elements. ~Fortunately, it appears we can workaround this by also wrapping the trigger and content into custom elements, which will prevent the browser from pulling them out of the `<p>` tag.~

cpsievert:
Closes #720 

The actual fix here is to switch from a `<div style="display:none">` to a `<template>` for the content "wrapper" elements, which will prevent the browser from pulling out that wrapper element out when it appears within a `<p>` tag (then, in the element is connected, we replace the `<template>` with `<div style="display:none">`).

While I was here, I realized that we don't need to set

```js
createRenderRoot(): this {
    return this;
  }
```

on our base element since (almost) all the markup is being provided as children to the custom element (and hence is treated as light DOM). Also, although we do construct markup for the close button inside the custom element (shadow DOM), when it gets shown, it gets pulled out of shadow DOM into the light DOM.


gadenbuie (approved):
Looks great, using `<template>` and slots seems like a solid approach.

PR #730, opened 2023-08-04
Be more careful about finding/creating tooltip trigger element

### LINKED ISSUE #728: Tooltip shows then immediately hides after repeatedly hovering a tooltip
For a minimal reprex:

```r
tooltip("I'm a tooltip", "A tip")
```

Try hovering in and out. After a couple hovers, hovering briefly shows the tooltip but then it immediately gets hidden

cpsievert:
Closes #728. 

The problem that leads to #728 is that if `this.triggerElement` (i.e., `getOrCreateTriggerEl()`) is requested _when the tooltip is shown_, then the contents won't be a child of the custom element (thus, breaking one of the assumptions of the `getOrCreateTriggerEl()` logic). This fix works around that by finding an immediate child with a `data-bs-toggle='tooltip/popover'` attribute (and taking that as the trigger)


cpsievert:
Sorry to merge this before a proper review @gadenbuie, but I'd like to get this change covered in Monday's testing results.

Feel free to comment or discuss next week

PR #814, opened 2023-10-02
More polish for shiny preset

cpsievert:
This PR currently does 5 things:

* On desktop, increases right padding (of the main content area) to match the left padding
* On mobile, increases top padding (of the main content area) and decreases left padding (to match the right padding)
* Removes underline styling from navbar links on mobile
* Increases the collapse breakpoint for navbars to lg devices
* Center aligns the navbar collapse icon in the navbar

These differences are probably best seen demo'd with the flights app: `shiny::runApp("inst/examples/flights")`

### TODO

- [ ] Maybe the last 2 should be pulled out into separate PRs since they are more general than the preset?
- [ ] It is worthwhile to bottom align `.nav-underline` inside `.navbar`?
- [ ] Do we need to be adding `.navbar-expand-lg` to `.navbar` in other projects?
  * Maybe this is a good enough reason to use an `@extend` here? 
- [ ] Make 2nd tab of flights demo looks better / more interesting


cpsievert:
Closing this in favor of smaller, bite-size, PRs

PR #823, opened 2023-10-05
`save_html()` now works correctly with components and pages with non-default theme values

### LINKED ISSUE #477: `card()` can't be saved directly to HTML with `save_html()`
Essentially because it works via `as_fragment()` which defines a custom print method

```r
htmltools::save_html(bslib::card("Hello"), "index.html")
```

cpsievert:
Closes #477

### TODO

- [ ] Maybe some snapshot tests would be worthwhile here?


DESCRIPTION
```diff
@@ -32,7 +32,7 @@ Imports:
     base64enc,
     cachem,
     grDevices,
-    htmltools (>= 0.5.4.9000),
+    htmltools (>= 0.5.6.9001),
```
cpsievert, line 35: `main` should already be on `0.5.6.9000` 🙈

gadenbuie (approved):
Nice!

PR #845, opened 2023-10-17
Require larger device size to expand navbar toggle; vertically center align toggle icon

cpsievert:
With this PR, 

* The collapse toggle for collapsible navbars (e.g., `page_navbar(collapsible = T)` and `navset_bar(collapsible = T)`) will now apply at the `lg` instead of `sm` breakpoint. 
* The collapse toggle (as well as the `title` is now guaranteed to be vertically centered)

For example, `shiny::runApp("inst/examples/flights")`:

### Before
Screenshot 2023-10-17 at 5 50 44 PM

### After
Screenshot 2023-10-17 at 5 51 23 PM


gadenbuie (approved):

cpsievert:
For posterity, I agree 100% with what you’ve said, but the unfortunate reality is moving things in the “right” direction (e.g., getting rid of the awkward `@extend` and/or moving away from `.navbar-header`, which doesn't exist anymore) would be a fair amount of work (not only in bslib, but rmarkdown, flexdashboard, etc), and we’re not really making the current situation worse off with this change

PR #852, opened 2023-10-19
Remove setting to window.bslib inside card.ts

schloerke:
Already performed in webcomponents/index.ts


schloerke:
Misunderstood. Closing

PR #855, opened 2023-10-20
Update version in `value_box()`'s deprecation method

cpsievert:
This should cover all relevant occurrences of `0.5.2` and `deprecate_`


gadenbuie (approved):
Thanks!

PR #857, opened 2023-10-23
docs: Use `a11y` arg from docs for tooltip example

schloerke:
Related docs: https://github.com/rstudio/bslib/pull/825/files#diff-7e26871f21e46dd166aac31c26b3b59c7a956dbc64a9ed980d2b8509165ec3bbR17

- [x] Docs are built


gadenbuie (approved):
Thanks!

PR #874, opened 2023-10-28
Change `AccordionInputBinding.receiveMessage` to async

### LINKED ISSUE #872: Unable to open a newly inserted accordion panel
This issue was introduced by `Shiny.renderContent()` becoming async (in https://github.com/rstudio/shiny/pull/3904).

In hindsight, it's quite obviously a bad idea for `Shiny.renderContent()` to always be async since, with any `insert`-like operation, users are going to expect that the operation to have finished before the next line of R code happens. Here's an example of that:

<details>
<pre><code>app.R</code></pre>

```r

remotes::install_github("rstudio/shiny")

library(bslib)
library(shiny)

ui <- page_fixed(
  accordion(
    id = "acc", multiple = TRUE,
    accordion_panel("A", "a"),
    accordion_panel("B", "b")
  )
)

  observe({
    # Triggers a Shiny.renderContent()
    accordion_panel_insert(
      "acc", accordion_panel("C", "c")
    )
    # Relies on the content being rendered
    accordion_panel_open("acc", "C")
  })
  
}

shinyApp(ui, server)
```
</details>

Going forward, it seems like it'd be a lot better to revert [changes made to `srcts/src/shiny/render.ts`](https://github.com/rstudio/shiny/pull/3904/files#diff-7aa61053d5a807425314d89085e38603c93b1c63a841d378e279aa3cd70417de
) (except for the `await`s in `renderContentAsync()`).

wch:
This fixes #872. It should be used with https://github.com/rstudio/shiny/pull/3930.

With this PR, this what the test app from #872 looks like when it starts:

image


wch:
Note that I wasn't able to successfully re-run `yarn` because I have a different system version of yarn from what others are using. (Once #788 is merged then we'll be able to consistently run `yarn` without worrying about the system version.)

Before releasing, we'll want to update the shiny version referred to in DESCRIPTION and in package.json.

wch:
Just to add a bit of detail: each message sent (via the R functions `accordion_panel_insert()` and `accordion_panel_open()` ends up putting a message on the JS side in the `actionQueue`, where they are handled as serially -- by `await`ing each message handler.

This PR makes `receiveMessage` async, and within it, it `await`s each call to `renderContent`. These changes (along with the PR in Shiny) ensure that the messages are handled in order, without any interleaved processing of them.

srcts/src/components/accordion.ts
```diff
@@ -141,13 +141,13 @@ class AccordionInputBinding extends InputBinding {
 
     // If there is still no targetItem, then there are no items in the accordion
     if (targetItem) {
-      Shiny.renderContent(
+      await Shiny.renderContent(
```
wch, line 144: Change to `renderContentAsync`

PR #993, opened 2024-02-26
Undo margin-bottom where appropiate when the parent of a dynamic UI slot is a gap-spacing context

### LINKED ISSUE #992: Extra unwanted spacing when using dynamic UI to render cards/value boxes/etc in gap context
Note that this example yields "double padding" between the cards, since the margin bottom of the 1st card fails to get undone:

```r
library(bslib)

ui <- page_fillable(
  uiOutput("card_1"),
  card("card 2")
)

  output$card_1 <- renderUI(card("card 1"))
}

shinyApp(ui, server)
```

The `margin-bottom` for the 1st card should get set to 0 like it does in this example:

```r
library(bslib)

page_fillable(
  card("card 1"),
  card("card 2")
)
```

cpsievert:
Closes #992


gadenbuie (approved):
Nice!

PR #1054, opened 2024-05-24
Shiny type update

wch:
This updates the types for rstudio/shiny to use the version from https://github.com/rstudio/shiny/pull/4063.

Note that this required a change to some types in `_utils.ts`, but this does not affect any generated JS files.

- [x] After the Shiny PR is merged, update package.json her to use `rstudio/shiny#main` and re-run `yarn` to update yarn.lock.


gadenbuie (approved):
Thanks!

PR #1065, opened 2024-06-03
fix documentation formatting

daattali:


R/sidebar.R
```diff
@@ -14,7 +14,7 @@
 #' * [page_sidebar()] creates a "page-level" sidebar.
 #' * [page_navbar()] creates a multi-panel app with an (optional, page-level)
 #'   sidebar that is shown on every panel.
-#' * `layout_sidebar()` creates a "floating" sidebar layout component which can
+#' * [layout_sidebar()] creates a "floating" sidebar layout component which can
```
cpsievert, line 17: This would just link back to the same page, so I prefer the current approach of no link

PR #1093, opened 2024-07-24
Update news; update example app in response to card_image() changes

cpsievert:
Update NEWS in preparation for release.

Also updates `inst/examples-shiny/card/app.R` in response to changes in `card_image()` to keep its behavior consistent (we use it for testing)


gadenbuie (approved):
Thanks!

PR #1204, opened 2025-04-25
feat(input_submit): Add new `input_submit_textarea()`/`input_submit_button()` components

cpsievert:
Work in progress that will depend on incoming changes to Shiny.

If anyone is curious to see examples, here are a couple basic ones:

<details>
<summary>app-submit-text.R</summary>

```r
library(shiny)
library(bslib)

ui <- page_sidebar(
  title = "Hello input_submit_textarea()",
  input_dark_mode(),
  input_submit_textarea(
    "text",
    "Enter input to submit",
    label = tags$code("bslib::input_submit_textarea()"),
  ),
  textOutput("value"),
  textAreaInput(
    "update_text",
    tags$code("shiny::textAreaInput()"),
    placeholder = "Enter text for updating the input above",
    autoresize = TRUE
  ),
  actionButton("update", "Update"),
  actionButton("focus", "Update & Focus"),
  actionButton("submit", "Submit"),
  actionButton("update_label", "Update Label"),
  shiny::useBusyIndicators(),
  sidebar = sidebar(open = FALSE),
  class = "align-items-center"
)

  output$value <- renderText({
    req(input$text)
    Sys.sleep(3)
    paste("You entered:", input$text)
  })

  observeEvent(input$update, {
    update_submit_textarea("text", value = input$update_text)
  })

  observeEvent(input$focus, {
    update_submit_textarea("text", value = input$update_text, focus = TRUE)
  })

  observeEvent(input$submit, {
    update_submit_textarea("text", value = input$update_text, submit = TRUE)
  })

  observeEvent(input$update_label, {
    update_submit_textarea("text", label = input$update_text)
  })

}

shinyApp(ui, server)
```

</details>

<details>
<summary>app-submit-button.R</summary>

```r
library(shiny)
library(bslib)

ui <- page_sidebar(
  title = "Hello input_submit_button()",
  sidebar = sidebar(
    sliderInput("n", "A number", 1, 100, 10),
    selectInput("letter", "A letter", LETTERS),
    input_submit_button(
      "submit",
      "Submit",
      icon = icon("paper-plane")
    ),
    helpText(
      "Changes to sidebar inputs aren't sent to the server until the submit button is clicked.",
      class = "lead"
    )
  ),
  card(
    card_header("Sidebar input values"),
    verbatimTextOutput("n_out"),
    verbatimTextOutput("letter_out"),
  ),
  card(
    card_header("Other input values"),
    layout_sidebar(
      sidebar = selectInput("state", "A state", state.abb),
      verbatimTextOutput("state_out"),
    )
  ),
  useBusyIndicators()
)

  output$n_out <- renderText({
    #req(input$submit)
    Sys.sleep(2)
    paste("Number:", input$n)
  })

  output$letter_out <- renderText({
    #req(input$submit)
    Sys.sleep(2)
    paste("Letter:", input$letter)
  })

  output$state_out <- renderText({
    paste("State:", input$state)
  })

}
shinyApp(ui, server)
```

</details>


srcts/src/components/submitTextArea.ts
```diff
@@ -0,0 +1,193 @@
...
+const EVENT_NAMESPACE = "textSubmitInputBinding";
+
+class TextAreaSubmitInputBinding extends InputBinding {
+  #submitButton: HTMLButtonElement | null = null;
+
+  find(scope: HTMLElement): JQuery<HTMLElement> {
+    return $(scope).find(".bslib-input-textsubmit > textarea");
+  }
+
+  initialize(el: HTMLTextAreaElement): void {
```
cpsievert, line 22: Make sure to update height in a similar way to `TextAreaInputBinding` does now https://github.com/rstudio/shiny/pull/4210

DESCRIPTION
```diff
@@ -51,14 +51,14 @@ Suggests:
     magrittr,
     rappdirs,
     rmarkdown (>= 2.7),
-    shiny (> 1.8.1),
+    shiny (> 1.11.0),
```
cpsievert:
```suggestion
    shiny (> 1.11.1),
```

R/utils-shiny.R
```diff
@@ -64,3 +64,15 @@ anyNamed <- function(x) {
...
+# Copy of shiny:::shinyInputLabel()
+shiny_input_label <- function(id, label = NULL) {
+  tags$label(
+    label,
+    class = "control-label",
+    class = if (is.null(label)) "shiny-label-null",
+    id = paste0(id, "-label"),
+    `for` = id
+  )
+}
```
gadenbuie: If we're only needing to duplicate the logic here, maybe just reach into shiny's namespace. There's such a tiny chance we'd remember to update this function if we make changes in shiny.

```suggestion
shiny_input_label <- function(id, label = NULL) {
  asNamespace("shiny")[["shinyInputLabel"]](id, label)
}
```

srcts/src/components/submitTextArea.ts
```diff
@@ -0,0 +1,209 @@
...
+    const btn = el.nextElementSibling;
+    // This assumption is forced server-side
+    if (!(btn instanceof HTMLButtonElement)) {
+      throw new Error("No submit button found");
+    }
```
gadenbuie: I'd rather we add a class to the button. These positional assumptions always make me nervous and it unnecessarily couples the R and JS logic, which makes it more likely that we break something by accident if we do any kind of moving things around.

The class approach is also nice because it gives users something direct to work with to style the text input submit button rather than also relying on positional selectors that we could accidentally break in the future.

R/input-submit.R
```diff
@@ -0,0 +1,161 @@
...
+  if (is.null(button)) {
+    button <- input_task_button(
+      id = paste0(id, "_submit"),
+      class = "btn-sm",
+      label = "Submit \U23CE",
+      title = "Press Enter to Submit",
+      `aria-label` = "Press Enter to Submit",
+      # The label/title above are updated in JS if a modifier key is needed
+      # (since it's platform dependent)
+      `data-default-button` = ""
```
gadenbuie: Is this attribute still needed or is it a vestige of the submit button?

R/input-submit.R
```diff
@@ -0,0 +1,161 @@
...
+#' @export
+input_submit_textarea <- function(
+  id,
+  label = NULL,
+  placeholder = NULL,
+  value = "",
+  ...,
+  button = NULL,
+  width = "min(680px, 100%)",
+  submit_key = c("enter", "enter+modifier")
```
gadenbuie: Dropping a pin here to say that I prefer `enter+modifier` as the default, but recognize that might not be the best option overall.

R/input-submit.R
```diff
@@ -0,0 +1,161 @@
...
+#'     Sys.sleep(2)
+#'     paste("You entered:", input$text)
+#'   })
+#' }
+#' shinyApp(ui, server)
+#'
+#' @section Server value:
+#' A character string of the text input. The default value is `""` even if
+#' `value` is provided. The value will only be set/updated when the user submits
+#' the input by pressing the Enter key or clicking the submit button.
```
gadenbuie: Small wording adjustments because I was initially confused by "the default value is `""`".

```suggestion
#' @section Server value:
#' The server receives a character string containing the user's text input.
#'
#' **Important:** The initial server value is always `""` (empty string),
#' regardless of any `value` parameter provided to `input_submit_textarea()`.
#' The server value updates only when the user explicitly submits the input by
#' either pressing the Enter key (possibly with a modifier key) or clicking the
#' submit button.
```

R/input-submit.R
gadenbuie: As I wend to submit my review, I noticed that GitHub (or maybe refined github) adds the submit key to the button label.

Image

I think it'd be great if we did this too, or did something similar through another affordance

inst/components/scss/input_submit_textarea.scss
```diff
@@ -0,0 +1,40 @@
+$prefix: bs- !default;
+
+.bslib-input-textsubmit {
+  display: flex;
+  flex-direction: column;
+  border: var(--#{$prefix}border-width, 1px) solid var(--#{$prefix}gray-500, #ced4da); // bs_theme(preset="shiny") sets input border color to $gray-500
+  border-radius: var(--#{$prefix}border-radius-lg, 4px);
```
gadenbuie: What do you think about using `-sm` instead (which is 4px by default, at least in the Shiny preset)?

```suggestion
  border-radius: var(--#{$prefix}border-radius-sm, 4px);
```

My first reaction when I saw the input was that the borders might feel too rounded compared with other components in their default styles

**-lg**

Image

**-sm**

Image

R/input-submit.R
```diff
@@ -0,0 +1,161 @@
...
+  label = NULL,
+  placeholder = NULL,
+  value = "",
+  ...,
+  button = NULL,
+  width = "min(680px, 100%)",
+  submit_key = c("enter", "enter+modifier")
+) {
+  rlang::check_installed("shiny", version = "1.11.1")
+  rlang::check_dots_empty()
```
gadenbuie: I just realized we're not making it easy to add classes or styles anywhere in this component. Suppose, e.g. I want a tighter border radius on the text input, or to adjust the font in the text input component.

Since users can provide the button component as an argument, it might be nice to allow attributes in `...` to be passed to `tags$textarea()`

R/input-submit.R
```diff
@@ -0,0 +1,161 @@
...
+        class = "form-control",
+        style = css(width = if (!is.null(width)) "100%"),
+        placeholder = placeholder,
+        `data-needs-modifier` = if (needs_modifier) "",
+        rows = 1,
+        value
+      ),
+      button
+    )
+  )
```
gadenbuie, line 143: I'm curious if we could flatten this component one level, i.e. something like this

```
.bslib-input-textsubmit .shiny-input-container
  label
  textarea
  button
```

If that's possible, it'd be my preference as it'd make the structure a little easier. I guess in particular I don't like that `.bslib-input-textsubmit` is the sort-of key identifying class, but it's not at the top of the component, so you'd need something like `.shiny-input-container:has(> .bslib-input-text-submit)` to be able to find the component "root"

R/input-submit.R
gadenbuie: It already does this! Awesome 😄 

image

R/input-submit.R
gadenbuie: Oh! It's so that we only add the submit key hint to our own default button.

I think instead, we should include a `<span class="submit-key">` in the button label and then update that span directly, which would make it easier both for custom buttons to get the correct submit key and to style the key hint.

srcts/src/components/submitTextArea.ts
cpsievert: And if a user supplies their own button, we `tagAppendAttributes()` to it?

R/input-submit.R
cpsievert: > Is this attribute still needed or is it a vestige of the submit button?

It's needed as a flag for the client-side logic to add a platform-specific modifier key symbol

R/input-submit.R
cpsievert: Yea, the thing pulling me towards `"enter"` as a default is for consistency with `chat_ui()`

R/input-submit.R
```diff
@@ -0,0 +1,161 @@
...
+  value <- shiny::restoreInput(id = id, default = value)
+  if (length(value) != 1 || !is.character(value)) {
+    stop("`value` must be a character string", call. = FALSE)
+  }
+
+  submit_key <- rlang::arg_match(submit_key)
+  needs_modifier <- isTRUE(submit_key == "enter+modifier")
+
+  if (is.null(button)) {
+    button <- input_task_button(
```
gadenbuie, line 97: Is the task button supposed to go to a spinner when you submit and while work is happening? It's not for me with the demo example and I'm not sure why not.

inst/components/scss/input_submit_textarea.scss
cpsievert: I was pulled towards `-lg` for consistency with the chat input, but it's probably better to align this with other inputs in mind 👍

R/input-submit.R
cpsievert: Yes, that is alarming -- do you have the latest shiny?

srcts/src/components/submitTextArea.ts
```diff
@@ -0,0 +1,209 @@
...
+        // If a modifier is needed, and it is present, submit the input
+        if (needsModifier && hasModifier) {
+          event.preventDefault();
+          btn.click();
+          return;
+        }
+
+        // If no modifier is needed, shift+enter inserts a new line,
+        // but enter alone submits the input
+        if (!needsModifier && !event.shiftKey) {
```
gadenbuie: I don't think this shows up in the `input_submit_textarea()` docs, but it'd be nice to mention somewhere. Also, I'm realizing that I often use `Alt+Enter` as well for new lines, so maybe also worth supporting that?

R/input-submit.R
gadenbuie: Oof, I mean I'm ambivalent about the default in `input_submit_textarea()` but I feel pretty strongly that it's the wrong behavior in `chat_ui()`

R/input-submit.R
```diff
@@ -0,0 +1,161 @@
...
+#' }
+#' shinyApp(ui, server)
+#'
+#' @section Server value:
+#' A character string of the text input. The default value is `""` even if
+#' `value` is provided. The value will only be set/updated when the user submits
+#' the input by pressing the Enter key or clicking the submit button.
+#'
+#' @export
+input_submit_textarea <- function(
```
gadenbuie, line 67: So I know you were working on a submit button at the same time, but I wonder if this is the right name for this function. Imagine a world where we have `input_text()` and `input_textarea()`. Would it make sense for this to be `input_textarea_submit()` so that they hang together?

R/input-submit.R
gadenbuie: Gotcha. I like the idea of using a classed `<span>` for these instead because it would make it a lot easier to re-use in custom buttons or to adjust styles in the default case.

R/input-submit.R
gadenbuie: Oh! It's not working when `submit_key = "enter+modifier"` because you're changing the `btn.textContent` but the task button stores labels in slots, so you're clobbering the task button states.

I think the classed span idea would solve this problem too

R/input-submit.R
cpsievert: I've flopped on this a bit -- I think you could argue either way. I like that way for autocomplete reasons, but also `input_submit_textarea()` feels better aligned with `input_[action/bookmark/submit/task]_button()`

gadenbuie: Good point, yeah let's stick with `input_submit_textarea()`

R/input-submit.R
cpsievert: Wrong enough that we're going to change it? FWIW, the current behavior is consistent with ChatGPT, Claude, Discord, Slack, etc., so partially for that reason, it does feel like a good default. I get that the modifier key is nice for longer (multi-line) input, but I don't think we necessarily want to assume that'd always gonna be the case, especially something this abstract.

gadenbuie: lol yeah mostly wrong because I use `chat_ui()` most often in a coding context, but yeah broadly I can see it being ambiguous. It should definitely be configurable, so I guess my feelings of "wrongness" in this case is more about not being able to change the default.

R/input-submit.R
cpsievert: Should work now

R/input-submit.R
gadenbuie: One final, possibly zany thought: maybe `rows` is the signal. If you want 2+ rows, you want longer text input and `"enter+modifier"`; if you only have 1 row it's a simple chat interface and new lines aren't expected.

gadenbuie: I'm going to double down on this. I think `_textarea` signals that it's a longer text input and should be `"enter+modifier"` by default.

But that also implies there could be an `input_submit_text()` which is a variant of the same component but with a single line of input, i.e. `rows = 1`, that defaults to `"enter"` to submit. If we had `input_submit_text()`, there are some additional optimizations we could do, like using only an icon for the submit button with a tooltip to show the keycode and having the button inside the text input on the same line (i.e. like the current chat UI text input).

In fact, I think having separate `input_submit_text()` and `input_submit_textarea()` components is a great idea, in part because I've also been thinking about a button tray for the text area input and so it'll be worth having two separate components (mostly markup and css differences).

Image

gadenbuie: More evidence to back up that `<input type="text">` should be `Enter` to submit vs `<textarea>` should be `Enter + modifer`: https://stackoverflow.com/questions/57064691/is-implicit-submission-submit-on-enter-key-useful-to-people-using-readers-and

R/input-submit.R
cpsievert: The problem with flattening that way is then (AFAIK) it's not possible to put a border around the textarea & button. I suppose putting a button in the textarea could be a workaround, but I think I'd rather keep it as is

gadenbuie: > it's not possible to put a border around the textarea & button.

Why not? Because you want the `<label>` to be outside of the box? (Could you add a screenshot of that design to the PR description? And one without?)

My main comment here is that the primary class should be at the root of the component. If you can't flatten it, fine, but under the current structure it'd be extremely awkward to target labels of just this component, for example.

cpsievert: > Because you want the <label> to be outside of the box? 

Yes. This is how it would look inside:

Screenshot 2025-10-07 at 10 19 42 AM


Also [cdf4d8d](https://github.com/rstudio/bslib/pull/1204/commits/cdf4d8d82f76ed2868f8aa743cebd80deaf2a02f) moves the identifier class to the top-level

R/input-submit.R
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  submit_key <- rlang::arg_match(submit_key)
+  needs_modifier <- isTRUE(submit_key == "enter+modifier")
+
+  if (is.null(button)) {
+    button <- input_task_button(
+      id = paste0(id, "_submit"),
+      class = "btn-sm",
+      label = HTML("Submit <span class='modifier-key'></span> \U23CE"),
```
gadenbuie: I was thinking the initial value would be in the span, maybe just an oversight? Or am I missing something?

```suggestion
      label = HTML("Submit <span class='modifier-key'>\U23CE</span>"),
```

R/input-submit.R
cpsievert: After talking about this in person, we've landed on:

* `input_submit_textarea()` will default to `"enter+modifier"`.
* Punt on adding `input_submit_text()` for now, but leave open the possibility of adding later if it seems it's needed.
    * There is also the open question of whether it should be `<input type="text">` or `<textarea>` (i.e., akin to shinychat's input). The former has accessibility benefits, but makes it harder to review longer inputs.

R/input-submit.R
cpsievert: Addressed this as well as improved styles in [e5a41e4](https://github.com/rstudio/bslib/pull/1204/commits/e5a41e4b53a5a8d9179b7463f4ac3c61ca772fb8)

srcts/src/components/submitTextArea.ts
cpsievert: Addressed in [331aaba](https://github.com/rstudio/bslib/pull/1204/commits/331aabaec41f0d6b6fa145a8ec8d1b18cc7d738e) and [ff02c93](https://github.com/rstudio/bslib/pull/1204/commits/ff02c93011d341adb835124cb36bb7dd75474124)

R/input-submit.R
cpsievert: I ended up taking this in a direction that overlaps somewhat with the toolbar concept we discussed earlier ([e9ef2eb](https://github.com/rstudio/bslib/pull/1204/commits/e9ef2eba3037381e21bb27a856ad312a84c8d859)), so you can now do something like:

```r
input_submit_textarea(
  "text",
  "Your story",
  placeholder = "Once upon a time...",
  actionLink("about", "About")
)
```

to get

Screenshot 2025-10-07 at 5 12 10 PM

cpsievert:
@gadenbuie ready for another review pass when you are

R/input-submit.R
```diff
@@ -0,0 +1,180 @@
...
+#' keystroke, making it useful for chat boxes, comments, or other scenarios
+#' where users may compose and review their text before submitting.
+#'
+#' @param id The input ID.
+#' @param label The label to display above the input control. If `NULL`, no
+#'   label is displayed.
+#' @param ... Unnamed argument values: UI elements to include alongside the
+#'   submit button (e.g., help text, links, etc.). Named argument values:
+#'   additional attributes to apply to the underlying `<textarea>` element
+#'   (e.g., spellcheck, autocomplete, etc).
```
gadenbuie: I have reservations about passing unnamed arguments from `...` into the area next to the button.

For now, I'd rather we only accept named attributes and add them to the text area.

In general, I think this is a place where we shouldn't rely on `...` and should have an argument that takes the UI elements that would go into the tray, or tool bar or whatever we end up calling it. I think that not allowing children in `...` would give us some space to release bslib with `input_submit_textarea()` and figure out that API in the future.

My core objection is that I think we should be extremely reluctant to separate the destination of named and unnamed arguments in `...`. I think some separation is inevitable when it makes sense for attributes to be applied at a higher level in the DOM than where children are inserted, but they still need to be conceptually related to each other (things going into the sidebar, for example). 

In this case, where attributes go to `<textarea>` and children go into some other location, it seems too far away to me. Having an argument for the children destination makes the final solution more readable and understandable because it'll be clear where children end up by the name of the argument.

inst/components/scss/input_submit_textarea.scss
```diff
@@ -0,0 +1,59 @@
+$prefix: bs- !default;
+
+// Top-level container
+.bslib-submit-textarea {
+  margin: 0 auto;
+}
```
gadenbuie: I liked that this class had a `.bslib-input-` prefix, which follows `.bslib-input-switch` and `.bslib-input-dark-mode`

gadenbuie (commented):
I'm going to pull changes locally to poke around, but two things jumped out at me from the web view

R/input-submit.R
```diff
@@ -0,0 +1,180 @@
...
+  needs_modifier <- isTRUE(submit_key == "enter+modifier")
+
+  if (is.null(button)) {
+    button <- input_task_button(
+      id = paste0(id, "_submit"),
+      class = "btn-sm",
+      label = HTML("Submit <span class='bslib-submit-key'>\U23CE</span>"),
+      title = "Press Enter to Submit",
+      `aria-label` = "Press Enter to Submit"
+    )
```
gadenbuie, line 110: This looks a bit like a hack, but the task button puts `{icon}{label}` in the state slots. This suggestion below puts the spinner in the same place as the modifier key

```suggestion
    button <- input_task_button(
      id = paste0(id, "_submit"),
      class = "btn-sm",
      label = span(class = "bslib-submit-key", "\U23CE"),
      icon = "Submit",
      label_busy = div(
        class = "spinner-border spinner-border-sm ms-2",
        role = "status",
        span(class = "visually-hidden", "Processing...")
      ),
      icon_busy = "Submit",
      title = "Press Enter to Submit",
      `aria-label` = "Press Enter to Submit"
    )
```

Kapture 2025-10-08 at 12 47 23

(Note we could use `icon_busy = "Processing..."`, but I prefer the stable button appearance.)

inst/components/scss/input_submit_textarea.scss
cpsievert: With the increase in classes, I was feeling like the prefix was becoming too long and redundant. That said, I can see at least wanting the input container class to have it ( [138b83d](https://github.com/rstudio/bslib/pull/1204/commits/138b83d0c4fa660d24cfe6e13f351f1da15d9efd))

R/input-submit.R
cpsievert: Good point. I had briefly considered this, but hesitated on committing to name. Anyway, done in [7ccf8e8](https://github.com/rstudio/bslib/pull/1204/commits/7ccf8e8179489b16c22ff070e35442a86cd8a86d)

R/input-submit.R
```diff
@@ -0,0 +1,190 @@
...
+        `data-needs-modifier` = if (needs_modifier) "",
+        rows = rows,
+        ...,
+        value
+      ),
+      div(
+        class = "bslib-submit-btn-container",
+        div(toolbar, class = "bslib-submit-btn-toolbar"),
+        button
+      )
```
gadenbuie: Kind of awkward class names here, maybe this instead


```suggestion
      tags$footer(
        div(toolbar, class = "bslib-toolbar"),
        button
      )
```

And then use `.bslib-input-submit-textarea footer` and `.bslib-input-submit-textarea .bslib-toolbar` as the selectors in the Sass.

R/input-submit.R
```diff
@@ -0,0 +1,190 @@
...
+      title = "Press Enter to Submit",
+      `aria-label` = "Press Enter to Submit"
+    )
+  }
+
+  if (!is_button_tag(button)) {
+    abort("`button` must be a `tags$button()`")
+  }
+
+  button <- tagAppendAttributes(button, class = "bslib-submit-textarea-btn")
```
gadenbuie, line 117: Also add `.ms-auto` to push right when the footer wraps.

inst/components/scss/input_submit_textarea.scss
```diff
@@ -0,0 +1,66 @@
...
+  align-items: center;
+  gap: 0.5rem;
+}
+
+// Container for user-defined 'toolbar' items
+.bslib-submit-btn-toolbar {
+  display: flex;
+  gap: 0.25rem;
+  align-items: center;
+}
```
gadenbuie:
```suggestion
.bslib-input-submit-textarea footer {
  display: flex;
  flex-wrap: wrap;
  justify-content: space-between;
  align-items: center;
  gap: 0.5rem;
}

// Container for user-defined 'toolbar' items
.bslib-input-submit-textarea .bslib-toolbar {
  display: flex;
  align-items: center;
  gap: 0.25rem;
}
```

PR #1260, opened 2025-11-17
feat: Add toolbar input select

### LINKED ISSUE #1249: Toolbar: Select Input
This will create a single select suitable for use in `toolbar`s.

Image

**Decisions:**

- It will allow users to select a *single* option from a drop-down. 
- It *will not* use Selectize.
- It will use the default menu open/close direction and formatting (vs. using selectize or adding custom JS callbacks to choose if it goes up or down)
- It will be borderless by default
- It will highlight on hover, focus, and selection
- Text will take the inherited size and match surrounding labels

elnelson575:
Fixes #1249 

Toolbar Input Selects should:
* Be single select with selectize=False
* Be slim enough to fit in a card header without altering the height
* Use the default behavior for dropdowns
* Have hover and select highlighting in light mode and in dark mode
* Accommodate all methods of specifying choices

**Decision Log:**
* Is it worth it to have a custom `selectOptions` if all we're doing is changing the error message? (Instead of importing from shiny)
    * Yes, make edits
* Background shading of input vs. border shadow? Background shading is more consistent with buttons I think.
    * Add outline shadow for selected state, make it consistent with the background/what is used with the buttons
* Do we need a custom input binding? What would that do? Yes, for update

Screenshot 2025-12-16 at 12 55 32 PM
Screenshot 2025-12-16 at 12 55 10 PM
Screenshot 2025-12-16 at 12 54 51 PM

<details>
<summary>Code</summary>

```
library(shiny)
library(bslib)

ui <- page_fillable(
  tags$head(
  ),
  card(
    card_header(
      "Card 1 header",
      toolbar(
        align = "right",
        gap = ".25rem",
        toolbar_input_button(
          id = "refresh",
          label = "Refresh",
          icon = icon("rotate"),
          tooltip = "Refresh data"
        ),
        toolbar_input_button(
          id = "download",
          label = "Download",
          icon = icon("download"),
          tooltip = "Download data"
        ),
        toolbar_input_select(
          id = "select",
          label = "Choose option",
          choices = c("Option 1", "Option 2", "Option 3"),
          selected = "Option 2",
          icon = icon("filter")
        ),
        toolbar_input_select(
          id = "letters",
          label = "Choose letter",
          choices = c("A", "B", "C"),
          selected = "B",
          class = "bg-success-subtle",
          tooltip = "Select a letter"
        ),
        input_dark_mode(id = "togglesm", class = "btn-xs border-0")
      )
    ),
    shiny::textOutput("selected_option"),
    p("Header toolbars:"),
    code(
      '
        toolbar_input_select(
          id = "select",
          label = "Choose option",
          choices = c("Option 1", "Option 2", "Option 3"),
          selected = "Option 2",
          icon = icon("filter")
        ),',
    ),
    code(
      '
        toolbar_input_select(
          id = "letters",
          label = "Choose letter",
          choices = c("A", "B", "C"),
          selected = "B",
          class = "bg-success-subtle",
          tooltip = "Select a letter"
        ),'
    ),
    p("Footer toolbar:"),
    code(
      '
         toolbar_input_select(
          id = "select_state",
          label = "Select state",
          choices = list(
            `East Coast` = list("NY", "NJ", "CT"),
            `West Coast` = list("WA", "OR", "CA"),
            `Midwest` = list("MN", "WI", "IA")
          ),
          selected = "NY",
          icon = icon("filter"),
          tooltip = "Choose a state from different regions of the US"
        )'
    ),
    max_height = "500px",
    card_footer(
      toolbar(
        align = "left",
        toolbar_input_select(
          id = "select_state",
          label = "Select state",
          choices = list(
            `East Coast` = list("NY", "NJ", "CT"),
            `West Coast` = list("WA", "OR", "CA"),
            `Midwest` = list("MN", "WI", "IA")
          ),
          selected = "NY",
          icon = icon("filter"),
          tooltip = "Choose a state from different regions of the US"
        )
      )
    )
  ),
  card(
    card_header(
      "Card 1 header",
      toolbar(
        align = "right",
        gap = ".25rem",
        toolbar_input_button(
          id = "refresh",
          label = "Refresh",
          icon = icon("rotate"),
          tooltip = "Refresh data"
        ),
        toolbar_input_button(
          id = "download",
          label = "Download",
          show_label = TRUE,
          icon = icon("download"),
          tooltip = "Download data"
        ),
        toolbar_input_select(
          id = "select",
          label = "Choose option",
          choices = c("Option 1", "Option 2", "Option 3"),
          selected = "Option 2",
          tooltip = "No icon, just text for choose option"
        ),
        toolbar_input_select(
          id = "letters",
          label = "Choose letter",
          choices = c("A", "B", "C"),
          selected = "B",
        ),
        input_dark_mode(id = "togglesm", class = "btn-xs border-0")
      )
    ),
    shiny::textOutput("selected_option"),
    p("Header toolbars:"),
    code(
      '
          toolbar_input_select(
          id = "select",
          label = "Choose option",
          choices = c("Option 1", "Option 2", "Option 3"),
          selected = "Option 2",
          tooltip = "No icon, just text for choose option"
        ),',
    ),
    code(
      '         toolbar_input_select(
          id = "letters",
          label = "Choose letter",
          choices = c("A", "B", "C"),
          selected = "B",
        ),'
    ),

    max_height = "500px"
  )
)

  output$selected_option <- renderText({
    paste("Selected option:", input$select)
  })

  observeEvent(input$select, {
    cat("Selected option:", input$select, "\n")
  })
}


```
</details>

### DIFF
--- a/tests/testthat/test-toolbar.R
+++ b/tests/testthat/test-toolbar.R
@@ -281,7 +281,7 @@ test_that("toolbar_input_select() rejects unnamed arguments in ...", {
   )
 })
 
-test_that("toolbar_input_select() has aria-labelledby with hidden label", {
+test_that("toolbar_input_select() has proper label structure", {
   tis <- as.tags(
     toolbar_input_select(
       id = "select",
@@ -291,33 +291,31 @@ test_that("toolbar_input_select() has aria-labelledby with hidden label", {
     )
   )
 
-  # The container is the root div element
-  container_elem <- tis
-  aria_labelledby <- htmltools::tagGetAttribute(
-    container_elem,
-    "aria-labelledby"
+  # Check that a <label> element exists with proper attributes
+  label_elem <- tagQuery(tis)$find("label")$selectedTags()[[1]]
+  expect_true(!is.null(label_elem))
+
+  # Label should have id matching pattern "{id}-label"
+  label_id <- htmltools::tagGetAttribute(label_elem, "id")
+  expect_equal(label_id, "select-label")
+
+  # Label should have for attribute pointing to select
+  label_for <- htmltools::tagGetAttribute(label_elem, "for")
+  expect_equal(label_for, "select")
+
+  # Find the label text span
+  label_spans <- tagQuery(label_elem)$find("span.bslib-toolbar-label")$selectedTags()
+  expect_true(length(label_spans) > 0)
+
+  label_text_span <- label_spans[[1]]
+  # With show_label=FALSE (default), label should be visually hidden
+  expect_match(
+    htmltools::tagGetAttribute(label_text_span, "class"),
+    "visually-hidden"
   )
 
-  # Check that aria-labelledby exists on container
-  expect_true(!is.null(aria_labelledby))
-
-  # Find the hidden label element by looking for spans with id that starts with "select-label-"
-  all_spans <- tagQuery(tis)$find("span")$selectedTags()
-  label_elem <- NULL
-  for (span in all_spans) {
-    span_id <- htmltools::tagGetAttribute(span, "id")
-    if (!is.null(span_id) && grepl("^select-label-", span_id)) {
-      label_elem <- span
-      break
-    }
-  }
-
-  expect_true(!is.null(label_elem))
-  label_id <- htmltools::tagGetAttribute(label_elem, "id")
-  label_text <- as.character(label_elem$children[[1]])
-
-  # Check that the hidden label has the correct id and text
-  expect_equal(label_id, aria_labelledby)
+  # Check label text content
+  label_text <- as.character(label_text_span$children[[1]])
   expect_equal(label_text, "Choose option")
 })
 


inst/components/scss/card.scss
```diff
@@ -31,6 +31,8 @@
     flex-direction: row;
     align-items: center;
     align-self: stretch;
+    min-height: 2.5rem;
```
elnelson575, line 34: This is only to match what we currently have in the buttons PR

inst/components/scss/toolbar.scss
```diff
@@ -2,35 +2,84 @@
 .bslib-toolbar {
   display: flex;
   align-items: center;
+  gap: 0;
```
elnelson575, line 12: Same with these changes to toolbar.scss for the .bslib-toolbar (Also just to match what's in buttons)

gadenbuie:
I'm not sure if the screenshot in the PR description reflects the current state, but my first reaction is that the horizontal padding feels imbalanced

image

inst/components/scss/card.scss
gadenbuie: note: this includes both `min-height` and `padding-block`

inst/components/scss/toolbar.scss
```diff
@@ -2,35 +2,84 @@
...
+
+  /* Make children take width=auto to size to content */
+  &>* {
+    width: auto;
+  }
+
+  select {
+    appearance: auto;
+    padding-block: 3px; // Decreased to slim down select to fit in header
+    /* .form-controls takes away the drop-down arrow, this ensures it shows up */
```
gadenbuie: I _think_ I understand what this comment is saying, but I'm not sure. Some things to clarify:

1. Which `.form-controls`? Do you mean Bootstrap's rule, Shiny's rule, or another rule in bslib?
2. I'm pretty sure `it` refers to the _drop-down arrow_, but I'm not sure what `this` refers to. I think you could probably re-order this sentence a bit to avoid two adjacent implicit references.

R/toolbar.R
```diff
@@ -26,3 +26,57 @@ toolbar <- function(
...
+#' @param choices List of values to select from. If elements of the list are
+#'   named, then that name - rather than the value - is displayed to the user.
+#'   It's also possible to group related inputs by providing a named list whose
+#'   elements are (either named or unnamed) lists, vectors, or factors. In this
+#'   case, the outermost names will be used as the group labels (leveraging the
+#'   `<optgroup>` HTML tag) for the elements in the respective sublist.
+#' @param selected The initially selected value. If not specified then defaults
+#'   to the first value.
+#' @param width The width of the input, e.g. '400px', or '100%'; see
+#'   [validateCssUnit()].
```
gadenbuie: This is probably sufficient; it looks like this documentation was copied from `selectInput()` anyway.

```suggestion
#' @inheritParams shiny::selectInput
```

gadenbuie: If you want to keep documentation local to `toolbar_input_select()` -- `width` might be one case -- you can keep the `@param` docs here.

R/toolbar.R
```diff
@@ -26,3 +26,57 @@ toolbar <- function(
...
+  id,
+  choices,
+  selected = NULL,
+  width = NULL
+) {
+  htmltools::div(
+    class = "bslib-toolbar-input-select form-select-sm",
+    shiny::selectInput(
+      id,
+      label = NULL, # Removed for slimline input component
```
gadenbuie: We probably need to keep `label` but use it for accessibility markup

R/toolbar.R
```diff
@@ -26,3 +26,57 @@ toolbar <- function(
...
+#' @family Toolbar components
+#' @export
+toolbar_input_select <- function(
+  id,
+  choices,
+  selected = NULL,
+  width = NULL
+) {
+  htmltools::div(
+    class = "bslib-toolbar-input-select form-select-sm",
```
gadenbuie: I think we might want to allow `...`, check that they're all named arguments (i.e. they are all attributes), and then let people add attributes to the outer container.


```suggestion
  width = NULL,
  ...
) {
  rlang::check_dots_named()
  
  htmltools::div(
    class = "bslib-toolbar-input-select form-select-sm",
    ...,
```

R/toolbar.R
```diff
@@ -26,3 +26,57 @@ toolbar <- function(
...
+    shiny::selectInput(
+      id,
+      label = NULL, # Removed for slimline input component
+      choices = choices,
+      selected = selected,
+      multiple = FALSE,
+      selectize = FALSE,
+      width = width
```
gadenbuie: `size = NULL` is the default for `selectInput()` and usually I wouldn't include it here, but if it's set you get a very different select input so it's worth setting this to ensure it doesn't accidentally change on us later.

```suggestion
      width = width,
      size = NULL
```

elnelson575:
> I'm not sure if the screenshot in the PR description reflects the current state, but my first reaction is that the horizontal padding feels imbalanced
> 
> image

@gadenbuie Some of that was because when there's sub groupings, it leaves room for the sub grouping header (i.e., the select input leaves room for the "East Coast" header even though "NY" is selected.
I def agree though that even the vanilla version looked a little off, so I played with that a bit.

Toolbar with no gap specified:
Screenshot 2025-12-04 at 10 25 21 AM

With a small gap specified in `toolbar()`:
Screenshot 2025-12-04 at 10 24 52 AM

How does that look? These do crunch closer to the arrow...too much?

R/toolbar.R
elnelson575: Removed width and the text is a little different (from shiny::selectInput) because don't want to reference the potential for multi select since we're not allowing that here.

R/toolbar.R
elnelson575: Yep-- agree. I took a stab at this and will push a good version. 


For reference, what we did for input button won't work here.
With:
```
  select_input <- shiny::selectInput(
    id,
    # We hide the label to make a slimmer input, but add an aria-label on the
    # select element for accessibility.
    label = span(hidden = NA, label),
    choices = choices,
    selected = selected,
    multiple = FALSE,
    selectize = FALSE,
    # Remove width as a direct arg to allow the container to flex, but it can
    # be set via the style attribute in `...`
    width = NULL,
    size = NULL
  )
```
and (to remove the blank space where the label would be created by the label div (created by shiny)
`  .control-label {
    display: none; // Hide label to make slimmer input
  }
`
Yields bad aria (where it says Card Header, it should say "Select Option", the text passed in via label):
input_select_hidden_label

R/toolbar.R
elnelson575: There is no `check_dots_named` function that I can find, only `check_dots_unnamed` so I followed a pattern used elsewhere in bslib (although there are a few different patterns used) to do the intention behind this. 
I added this, and removed width from the function args because that seemed like a relatively niche use case compared to auto-fit. If a standardized width is needed, it can be passed in here using `style = "width: 400px"`

R/toolbar.R
```diff
@@ -143,3 +143,82 @@ toolbar_input_button <- function(
...
+    size = NULL
+  )
+
+  # Add aria-label here instead of providing a hidden label in the selectInput
+  # because the hidden label still takes up space and even if entirely hidden
+  # using SCSS it is still better parsed by screen readers (when there is no
+  # visible text to reference) by using aria-label instead.
+  select_input <- tagQuery(select_input)$find("select")$addAttrs(
+    "aria-label" = label
+  )$allTags()
```
gadenbuie: Could we lean on `labelledby` as we did in the icon buttons? We don't have a `tagQuery` equivalent in Python, so it'd definitely better for us if we can avoid this kind of tag surgery.

R/toolbar.R
```diff
@@ -143,3 +143,82 @@ toolbar_input_button <- function(
...
+  label,
+  choices,
+  selected = NULL,
+  ...
+) {
+  # Validate that ... contains only named arguments
+  dots <- rlang::list2(...)
+  if (any(!nzchar(rlang::names2(dots)))) {
+    rlang::abort("All arguments in `...` must be named.")
+  }
```
gadenbuie, line 227: Sorry for leading you astray with `check_dots_named()`. I consistently forget which of the two doesn't exist. We'd clearly make use of a `check_dots_named()` variant but none has been established so far.

Anyway, you can use our internal helper `separate_arguments()`

```r
separate_arguments()
#> $attribs
#> list()
#> 
#> $children
#> list()
```

and then check that `$children` has length zero.


```suggestion
  dots <- separate_arguments(...)
  if (length(dots$children) > 0) {
    rlang::abort("All arguments in `...` must be named.")
  }
```

and use `!!!dots$attribs` below.

inst/components/scss/toolbar.scss
elnelson575: Thanks-- I expanded on this to make it clearer!

R/toolbar.R
elnelson575: So we could add a `labelledby`, but it would need to wrap the overall input vs. be applied to only the label. (If applied to only the label via a hidden span, it does not read the aria as associated with the select, I tried this with both `aria-labelledby` and `aria-label`).
Current pushed version now uses the wrapping strategy. It's mildly less optimal aria-wise from what I can tell because ideally the label is on the `select` itself, but based on my VoiceOver testing, it's still reading the correct aria with the updated strategy.

gadenbuie: After talking through all of the options, let's go with `toolbar_input_select()` using a bare-bones, self-contained implementation that creates it's own `<select>` input markup, which would result in something like this:

```html
<div class="bslib-toolbar-input-select shiny-input-container">
  <select id="select_state_manual" class="form-select form-select-sm border-0" aria-label="Default select example">
    <optgroup label="East Coast">
      <option value="NY" selected>NY</option>
      <option value="NJ">NJ</option>
      <option value="CT">CT</option>
    </optgroup>
    <optgroup label="West Coast">
      <option value="WA">WA</option>
      <option value="OR">OR</option>
      <option value="CA">CA</option>
    </optgroup>
    <optgroup label="Midwest">
      <option value="MN">MN</option>
      <option value="WI">WI</option>
      <option value="IA">IA</option>
    </optgroup>
  </select>
</div>
```

If you're comparing with `shiny::selectInput()`, we don't need many of the features: no selectize; no multiple selections; `label` can be enforced as a string because it's only used for `aria-label` (and must be provided!).

We **will** need to replicate the `choices` → list behavior, which should be relatively straight-forward. And we will need to call `restoreInput()` at the top of the component to enable bookmarking.

We may also find ourselves needing to create an input binding for this element so that we can disallow updating the select input to things that aren't possible with this select input. We can pair on that step once we have the basic markup in place.

R/toolbar.R
```diff
@@ -143,3 +143,128 @@ toolbar_input_button <- function(
...
+  perfWarning = FALSE
+) {
+  if (length(choices) >= 1000) {
+    warning(
+      "The select input \"",
+      inputId,
+      "\" contains a large number of ",
+      "options; this may cause performance issues.",
+      call. = FALSE
+    )
```
```suggestion
    rlang::warn(
      sprintf(
        "Select input `%s` contains a large number of option; this may cause performance issues.",
        inputId
      )
    )
```

R/toolbar.R
```diff
@@ -136,10 +150,216 @@ toolbar_input_button <- function(
...
+      container,
+      tooltip,
+      placement = "bottom"
+    )
+  }
+
+  container
+}
+
+# This function ported from shiny's `input-select.R` with minor changes.
```
gadenbuie: Can you please call out the minor changes in comments?

R/toolbar.R
```diff
@@ -136,10 +150,216 @@ toolbar_input_button <- function(
...
+#' @param id The input ID.
+#' @param label The label for the select input. A hidden label is created with
+#'   this text and referenced via `aria-labelledby` for accessibility. Must be
+#'   a non-empty string.
+#' @param choices List of values to select from. If elements of the list
+#'   are named, then that name — rather than the value — is displayed to
+#'   the user. It's also possible to group related inputs by providing a
+#'   named list whose elements are (either named or unnamed) lists,
+#'   vectors, or factors. In this case, the outermost names will be used
+#'   as the group labels (leveraging the ⁠<optgroup>⁠ HTML tag) for the
```
gadenbuie: Great description overall, but the use of `<optgroup>` is too in the weeds for the param docs

R/toolbar.R
```diff
@@ -136,10 +150,216 @@ toolbar_input_button <- function(
...
+#'   elements in the respective sublist.
+#' @param selected The initially selected value. If not provided, the first
+#'   choice will be selected by default.
+#' @param tooltip Tooltip text to display when hovering over the select
+#'   input. Can be:
+#'   * `NULL` (default) - no tooltip is shown
+#'   * A character string - wraps the entire input with a tooltip
```
gadenbuie: Since there are only two options, it'd be better to just use prose here. Some editors only show the first paragraph of text in hover tips and might miss the list, so I try to only use them when it's structurally important or easier to read.

gadenbuie: Also this behavior should be inline with what we've established for `toolbar_input_button(tooltip=)`

R/toolbar.R
elnelson575: This was ported from the original `selectInput` docs-- I took a stab at simplifying it in this next iteration, but lmk if you have ideas for further simplification.

gadenbuie: If we're just taking the descriptions directly from `shiny::selectInput()` we can use `@inheritParam shiny::selectInput`

inst/components/scss/toolbar.scss
```diff
@@ -92,3 +92,59 @@
     align-self: center;
   }
 }
+
+
+/* Toolbar Select Input */
+.bslib-toolbar-input-select {
+  padding-inline: 0.25rem;
+  height: var(--_icon-size, 1.75rem);
```
gadenbuie: `--_icon-size` doesn't exist anymore

R/toolbar.R
```diff
@@ -144,6 +144,215 @@ toolbar_input_button <- function(
...
+#'   align = "right",
+#'   toolbar_input_select(
+#'     id = "select",
+#'     label = "Choose option",
+#'     choices = c("Option 1", "Option 2", "Option 3"),
+#'     selected = "Option 2"
+#'   )
+#' )
+#'
+#' # With tooltip
```
```suggestion
#' # With custom tooltip
```

R/toolbar.R
```diff
@@ -144,6 +144,215 @@ toolbar_input_button <- function(
...
+#'
+#' @family Toolbar components
+#' @export
+toolbar_input_select <- function(
+  id,
+  label,
+  choices,
+  ...,
+  selected = NULL,
+  tooltip = FALSE,
```
```suggestion
  tooltip = TRUE,
```

R/toolbar.R
```diff
@@ -144,6 +144,215 @@ toolbar_input_button <- function(
...
+  icon = NULL
```
gadenbuie: We should consider adding `show_label` for consistency

R/toolbar.R
```diff
@@ -144,6 +144,215 @@ toolbar_input_button <- function(
...
+) {
+  # Import Shiny's internal choice processing functions
+  choicesWithNames <- asNamespace("shiny")[["choicesWithNames"]]
+  firstChoice <- asNamespace("shiny")[["firstChoice"]]
```
gadenbuie: Let's move these next to where they're used, e.g.

```r
firstChoice <- asNamespace("shiny")[["firstChoice"]]
selected <- firstChoice(choies)
```

R/toolbar.R
```diff
@@ -144,6 +144,215 @@ toolbar_input_button <- function(
...
+
+  # Restore input for bookmarking
+  selected <- shiny::restoreInput(id = id, default = selected)
+
+  # Set selected to the first choice if no default or restored value
+  if (is.null(selected)) {
+    selected <- firstChoice(choices)
+  }
+
+  # Validate that ... contains only named arguments
```
gadenbuie, line 223: Let's do all validation as early as possible in the function call

R/toolbar.R
gadenbuie: Because we're reaching into shiny's namespace, we should make sure that we have tests that will tell us immediately if these functions change in a fundamental way.

R/toolbar.R
```diff
@@ -144,6 +144,215 @@ toolbar_input_button <- function(
...
+# Create tags for each of the options; use <optgroup> if necessary.
+# This returns an HTML string instead of tags for performance.
+selectOptions <- function(
+  choices,
+  selected = NULL,
+  inputId,
+  perfWarning = FALSE
+) {
+  if (length(choices) >= 1000) {
+    rlang::warn(
```
gadenbuie, line 310:
```suggestion
    # This warning differs from `shiny:::selectOptions()` to remove
    # mention of sever-side options
    rlang::warn(
```

R/toolbar.R
```diff
@@ -144,6 +144,215 @@ toolbar_input_button <- function(
...
+      tooltip,
+      placement = "bottom"
+    )
+  }
+
+  container
+}
+
+# This function ported from shiny's `input-select.R` with changes to the
+# warning message.
```
```suggestion
# This function was copied from shiny's `input-select.R`
# with a wording change in the warning message
```

R/toolbar.R
gadenbuie: And add one test to compare our `selectOptions()` to `shiny:::selectOptions()`

R/toolbar.R
```diff
@@ -144,6 +144,215 @@ toolbar_input_button <- function(
...
+
+  # Add optional icon before the select
+  icon_elem <- NULL
+  if (!is.null(icon)) {
+    # Remove aria-label from icon if present. Even though the icon is wrapped
+    # in a span with aria-hidden="true", screen readers may still read the
+    # icon's aria-label. This ensures the icon is treated as purely decorative.
+    if (!is.null(icon$attribs$`aria-label`)) {
+      icon$attribs$`aria-label` <- NULL
+    }
```
gadenbuie: Maybe we do this with javascript instead? (can leave it here for now)

gadenbuie: Two other ideas:

* Set `role="none"` on the icon wrapper span
* Or set `aria-label=""` on the icon wrapper span

elnelson575: Unfortunately both of these led to a weird dual focus situation
Screenshot 2025-12-16 at 11 54 54 PM

gadenbuie: For posterity, the dual focus came from `tooltip()` being wrapped around a `div()`. Even though there's a focusable element inside the tooltip container, `<bslib-tooltip>` applies `tabindex="0"` to the first child so that it's reachable by keyboard (we could maybe be a little smarter about this? or this is an acceptable edge case?). Anyway, we moved the tooltip onto the `<select>` element so that focusing the tooltip also causes the select element to be focused.

gadenbuie (approved):
🎉 🔽 ✅

PR #1265, opened 2025-12-08
feat: Add toolbar spacer

### LINKED ISSUE #1259: Toolbar: Add toolbar spacer
Toolbar could use a `toolbar_spacer()` which could be a space, dividing line OR a nav_spacer() style "push the things over".

For example:

Image

elnelson575:
Fixes #1259 

**Goal:**
Give users control over spacing in toolbars.
_Enable both:_
* Dividers with lines
* Spacers that are blank

_Allow users to:_
* Change width of blank spacer
* Change width of spacer with divider (via the spacing around the line)
* Change width of divider line (via function arg)
* Change color of divider line (via bootstrap var, defaults to the border color)

_Additional Considerations:_
* Should not read anything in VoiceOver (verified)
* Should work if switched to dark mode
<details>
<summary>Dark Mode Verification </summary>
Screenshot 2025-12-08 at 11 06 43 AM
</details>

**Decision Log:**
_How to we want to make a toolbar like this particular (split) case?_
Screenshot 2025-12-05 at 12 43 04 PM
This can be accomplished right now via creating two toolbars, one with align = "left" and one with align = "right".
Question: Should we make it possible such that you have ONE toolbar (vs two with different alignments) and put a spacer in the middle to split half of it to the left and half of it to the right? 
**No.** In the future, we could consider creating an `align="stretch"` if we wanted to enable this, but for now, let's just go with the two toolbars.

**Example App:**
Screenshot 2025-12-10 at 3 42 39 PM

<details>
<summary>Code</summary>

```
library(shiny)
library(bslib)

ui <- page_fillable(
  h2("Toolbar Divider Examples"),

  layout_columns(
    # Example 1: Divider and gap widths
    card(
      card_header(
        "Dividers (Default and Custom Width)",
        toolbar(
          toolbar_input_button(
            id = "btn1",
            label = "Home",
            icon = icon("home")
          ),
          toolbar_divider(), # Default divider
          toolbar_input_button(
            id = "btn2",
            label = "Settings",
            icon = icon("gear")
          ),
          toolbar_divider(gap = "4rem"), # Custom gap
          toolbar_input_button(
            id = "btn3",
            label = "Notifications",
            icon = icon("bell"),
          ),
          toolbar_divider(width = "4px"), # Custom divider width
          toolbar_input_button(
            id = "btn4",
            label = "Profile",
            icon = icon("user")
          ),
          # Custom divider width and gap
          toolbar_divider(width = "4px", gap = "4rem")
        )
      ),
      pre(code(
        "toolbar(
          toolbar_input_button('btn1', 'Home', icon('home')),
          toolbar_divider(), # Default divider
          toolbar_input_button('btn2', 'Settings', icon('gear')),
          toolbar_divider(gap = '4rem'), # Custom gap
          toolbar_input_button('btn3', 'Notifications', icon('bell')),
          toolbar_divider(width = '4px'), # Custom divider width
          toolbar_input_button('btn4', 'Profile', icon('user')),
          # Custom divider width and gap
          toolbar_divider(width = '4px', gap = '4rem')
        )"
      )),
    ),

    # Example 2: And now in a footer
    card(
      card_header(
        "All the excitement is in the footer",
      ),
      pre(code(
        "toolbar(
          toolbar_input_button('btn1f', 'Home', icon('home')),
          toolbar_divider(),
          toolbar_input_button('btn2f', 'Settings', icon('gear')),
          toolbar_divider(gap = '4rem'),
          toolbar_input_button('btn3f', 'Notifications', icon('bell')),
          toolbar_divider(width = '4px'),
          toolbar_input_button('btn4f', 'Profile', icon('user')),
          toolbar_divider(width = '4px', gap = '4rem')
)"
      )),
      card_footer(
        toolbar(
          toolbar_input_button(
            id = "btn1f",
            label = "Home",
            icon = icon("home")
          ),
          toolbar_divider(), # Default divider
          toolbar_input_button(
            id = "btn2f",
            label = "Settings",
            icon = icon("gear")
          ),
          toolbar_divider(gap = "4rem"), # Custom gap
          toolbar_input_button(
            id = "btn3f",
            label = "Notifications",
            icon = icon("bell")
          ),
          toolbar_divider(width = "4px"), # Custom divider width
          toolbar_input_button(
            id = "btn4f",
            label = "Profile",
            icon = icon("user")
          ),
          # Custom divider width and gap
          toolbar_divider(width = "4px", gap = "4rem")
        )
      )
    ),
  ),

  # Example 3: Split toolbars and gap spacing
  card(
    card_header(
      toolbar(
        align = "left",
        toolbar_input_button(
          id = "add",
          label = "Add",
          icon = icon("plus"),
        ),
        toolbar_divider(),
        toolbar_input_button(
          id = "remove",
          label = "Remove",
          icon = icon("minus"),
        )
      ),
      toolbar(
        align = "right",
        toolbar_input_button(id = "save", label = "Save", icon = icon("save")),
        toolbar_divider(width = "0px", gap = "2rem"),
        toolbar_input_button(
          id = "edit",
          label = "Edit",
          icon = icon("pencil")
        ),
        toolbar_divider(width = "0px", gap = "2rem"),
        toolbar_input_button(
          id = "delete",
          label = "Delete",
          icon = icon("trash")
        )
      )
    ),
    p("This example shows split toolbars with gap spacing between buttons."),
    pre(code(
      "toolbar(
        align = 'right',
        gap = '0.5rem',
        toolbar_input_button(id = 'save', label = 'Save', icon = icon('save')),
        toolbar_divider(width = '0px', gap = '2rem'),
        toolbar_input_button(id = 'edit', label = 'Edit', icon = icon('pencil')),
        toolbar_divider(width = '0px', gap = '2rem'),
        toolbar_input_button(id = 'delete', label = 'Delete', icon = icon('trash'))
      )"
    ))
  )
)

```

</details>

**Remaining To-Do:**
- [x] Check unit test coverage
- [x] Add an entry to NEWS.md concisely describing what you changed.

### DIFF
--- a/inst/components/scss/toolbar.scss
+++ b/inst/components/scss/toolbar.scss
@@ -3,6 +3,7 @@
   --_divider-height: var(--bslib-toolbar-divider-height, 1lh);
   --_divider-width: var(--bslib-toolbar-divider-width, 2px);
   --_divider-gap: var(--bslib-toolbar-divider-gap, 1rem);
+  --_divider-color: var(--bslib-toolbar-divider-color, var(--bs-border-color-translucent, rgba(40,70,94,0.1)));
   // Sets the height/width (toolbar buttons are square) for toolbar buttons
   // Icon size is controlled by font-size
   --_toolbar-btn-size: var(--bslib-toolbar-btn-size, 1.75rem);
@@ -72,7 +73,7 @@
         display: block;
         width: var(--_divider-width);
         height: 100%;
-        background-color: var(--bs-border-color-translucent);
+        background-color: var(--_divider-color);
         margin: 0 auto;
       }
   }


R/toolbar.R
```diff
@@ -143,3 +143,59 @@ toolbar_input_button <- function(
...
+  }
+
+  as_fragment(div(
+    class = "bslib-toolbar-spacer",
+    # Adds the divider class to add the psudo-element for the divider line
+    class = if (has_divider) "bslib-toolbar-divider",
+    style = css(
+      width = width,
+      # Sets the width of the pseudo-element divider line if applicable
+      `--divider-width` = if (has_divider) divider_width else NULL
```
gadenbuie: Let's update this so that we're only setting the CSS variable when it's explicitly customized by the user, otherwise we should respect the defaults in the CSS. That will also open up a path for setting the default via CSS.

inst/components/scss/toolbar.scss
```diff
@@ -53,6 +53,27 @@
...
   }
 
+  /* ---- Toolbar Spacer ---- */
+
+  .bslib-toolbar-spacer {
+    align-self: stretch; // Ensures divider lines stretch to full height
+
+    // Divider variant: vertical line
+    &.bslib-toolbar-divider {
+      opacity: 0.75;
```
gadenbuie: I'm not sure I like the "full-height divider". I think it looks a little better when the divider is about as a tall as the button icon, or `1lh`.

```suggestion
  .bslib-toolbar-spacer {
    align-self: center;

    // Divider variant: vertical line
    &.bslib-toolbar-divider {
      opacity: 0.75;
      height: var(--_divider-height);
```

But we should also be adding this "internal CSS variables" at the component root, rather than deep in the component tree. I'd recommend something like this:

```scss
.bslib-toolbar {
  --_divider-height: var(--bslib-toolbar-divider-height, 1lh);
}
```

This both sets the default `--_divider-height` and creates a mechanism for setting the global default with CSS variables. A user can easily pick from the following:

```css
:root {
  --bslib-toolbar-divider-height, 1.25lh;
}

/* or in a specific subtree */
.my-card {
  --bslib-toolbar-divider-height: 1.25lh
}
```

or via R

```r
card(style = css("--bslib-toolbar-divider-height" = "1.25lh"))
```

etc. etc.

inst/components/scss/toolbar.scss
```diff
@@ -53,6 +53,27 @@
...
+
+      // Use a pseudo-element for the actual line so custom width controls spacing
+      &::before {
+        content: "";
+        display: block;
+        width: var(--divider-width, 2px);
```
gadenbuie: In the same vein as above, this should be `--_divider-width` and its default value should be configurable via a global CSS variable.

Note that internally in our R code we'll set `--_divider-width` and _not_ the global variable.

inst/components/scss/toolbar.scss
gadenbuie: Relatedly, we should clean up `--_icon-size` in this change too. We'll need a better name for that variable because it controls the button size, not specifically the icon size.

R/toolbar.R
```diff
@@ -143,3 +143,59 @@ toolbar_input_button <- function(
...
+    has_divider <- TRUE
+    divider_width <- validateCssUnit(divider)
+  }
+
+  as_fragment(div(
+    class = "bslib-toolbar-spacer",
+    # Adds the divider class to add the psudo-element for the divider line
+    class = if (has_divider) "bslib-toolbar-divider",
+    style = css(
+      width = width,
```
gadenbuie: I think we should probably do the same treatment for `width` as `divider` in the sense that we can have `width = NULL` as the default and only change the default width if specifically requested by the user. That would allow for global spacing changes to be made via CSS.

In this iteration, I think width should probably be second in the function

R/toolbar.R
```diff
@@ -143,3 +143,59 @@ toolbar_input_button <- function(
...
+#'   toolbar_input_button(id = "a", label = "A"),
+#'   toolbar_spacer(width = "20px", divider = "5px"),
+#'   toolbar_input_button(id = "b", label = "B"),
+#'   toolbar_spacer(divider = FALSE),
+#'   toolbar_input_button(id = "c", label = "C")
+#' )
+#'
+#' @family Toolbar components
+#' @export
+toolbar_spacer <- function(width = "1rem", divider = TRUE) {
```
gadenbuie: What do you think about renaming this `toolbar_divider()`? There's some conceptual overlap here with `nav_spacer()`, but they're also quite different in important ways.

I also think we could simplify organize the function around the divider and instead of overloading `divider = T/F`, we could have people set the divider width to 0 to hide it. Maybe something like this

```r
toolbar_divider()
toolbar_divider(width = "2px", gap = "1rem") # defaults used above
toolbar_divider(width = 0, gap = "2rem") # no dividing line
toolbar_divider(width = 4, gap = "0.5rem") # 4px dividing line, half size gap
```

We could use other words for `gap`, like `space` or something else. My instinct is that we should organize around the most prevalent use case, which is to add a dividing line. In that case, `width` seems better aligned with the divider. It's also the more obvious visual element and where precision counts. There's a big visual difference between 2px and 4px. This doesn't apply to the empty space as much, which is why `gap` or `space` (or padding/margin) would make sense to me here.

Anyway, this reorganization would also let us use a single `.bslib-toolbar-divider` class. TBF we can do that now, but it makes a lot more sense to have that be the primary/only class if we center the divider in the name.

elnelson575: As discussed, no qualms with `toolbar_divider()` as a name. `width` and `gap` are okay with me, as long as we make sure to be very clear when we're talking about divider line width vs. overall component width (in this case, gap) in the params and elsewhere.
Next version will have that when pushed!

R/toolbar.R
elnelson575: This is now passed as `--bs-toolbar-divider-width = width` where `width` defaults to `NULL` and the css sets `width` via `var(--bslib-toolbar-divider-width, 2px)`.
If `width` is `NULL`, the class is dropped, based on my testing. Lmk if you had an alternate approach in mind for this.

inst/components/scss/toolbar.scss
elnelson575: Did the above ✔️ 
Also, I updated `--_icon-size` to `--_toolbar_btn_size: var(--bslib-toolbar-btn-size, 1.75rem);` and then the `height` on the actual button to `height: var(--bslib-toolbar-btn-size, var(--_toolbar-btn-size));` with the thought that users could then make the buttons bigger if they wanted like:
```
          toolbar_input_button(
            id = "btn3",
            label = "Notifications",
            icon = icon("bell"),
            "style" = "--bslib-toolbar-btn-size: 2.5rem" # Custom button size
          ),
```

inst/components/scss/toolbar.scss
elnelson575: I assume you mean in our R code we'll set something like `--bslib-toolbar-divider-width` since `--_divider-width` would be the global?
Updated this based on what I thought you meant, let me know if I was incorrect.

R/toolbar.R
elnelson575: Updated this to use the `...` strategy mentioned in our chat.

inst/components/scss/toolbar.scss
```diff
@@ -1,5 +1,11 @@
 /* Toolbar */
 .bslib-toolbar {
+  --_divider-height: var(--bslib-toolbar-divider-height, 1lh);
+  --_divider-width: var(--bslib-toolbar-divider-width, 2px);
+  --_toolbar-gap: var(--bslib-toolbar-gap, 1rem);
```
gadenbuie: Let's call this `divider-gap` to not be confused with `.bslib-toolbar { gap: ... }`

inst/components/scss/toolbar.scss
```diff
@@ -53,6 +59,29 @@
     }
   }
 
+  /* ---- Toolbar Divider ---- */
+
+  .bslib-toolbar-divider{
+    align-self: center; // Ensures divider lines stretch to full height
```
```suggestion
    align-self: center;
```

inst/components/scss/toolbar.scss
```diff
@@ -53,6 +59,29 @@
...
+    // height)
+    width: var(--bslib-toolbar-gap, var(--_toolbar-gap));
+
+      // Use a pseudo-element for the actual line so custom width controls spacing
+      &::before {
+        content: "";
+        display: block;
+        width: var(--bslib-toolbar-divider-width, var(--_divider-width));
+        height: 100%;
+        background-color: var(--bs-border-color);
```
gadenbuie: Maybe we'd want `--bs-border-color-translucent` instead of border color + opacity.

R/toolbar.R
```diff
@@ -143,3 +143,47 @@ toolbar_input_button <- function(
 
   button
 }
+
+#' @describeIn toolbar Add a divider to a toolbar
```
```suggestion
#' Toolbar: Add a divider to a toolbar
```

R/toolbar.R
```diff
@@ -143,3 +143,47 @@ toolbar_input_button <- function(
...
+  as_fragment(div(
+    class = "bslib-toolbar-divider",
+    style = css(
+      # Sets the overall width of divider space
+      `--bslib-toolbar-gap` = gap,
+      # Sets the width of the pseudo-element divider line, defaults to 2px
+      `--bslib-toolbar-divider-width` = width
+    ),
+    `aria-hidden` = "true"
+  ))
```
```suggestion
  div(
    class = "bslib-toolbar-divider",
    style = css(
      # Sets the overall width of divider space
      `--_divider-gap` = gap,
      # Sets the width of the pseudo-element divider line, defaults to 2px
      `--_divider-width` = width
    ),
    `aria-hidden` = "true"
  )
```

No need for `as_fragment()` and we'll set the psuedo-private properties inline instead of the global custom properties

inst/components/scss/toolbar.scss
```diff
@@ -23,7 +29,7 @@
     align-items: center;
     justify-content: center;
     line-height: 1; // Override Bootstrap's line-height to avoid too much vertical space
-    height: var(--_icon-size, 1.75rem); // Keep square icon and label + icon button heights consistent
+    height: var(--bslib-toolbar-btn-size, var(--_toolbar-btn-size)); // Keep square icon and label + icon button heights consistent
```
```suggestion
    height: var(--_toolbar-btn-size); // Keep square icon and label + icon button heights consistent
```

gadenbuie: And in other places below, just make sure we're only using the `--_*` properties internally (and setting them at the outer edges of the component)

inst/components/scss/toolbar.scss
```diff
@@ -1,5 +1,11 @@
 /* Toolbar */
 .bslib-toolbar {
+  --_divider-height: var(--bslib-toolbar-divider-height, 1lh);
+  --_divider-width: var(--bslib-toolbar-divider-width, 2px);
+  --_divider-gap: var(--bslib-toolbar-divider-gap, 1rem);
```
gadenbuie, line 5: I think we should go ahead and follow the same pattern for the divider color

```suggestion
  --_divider-gap: var(--bslib-toolbar-divider-gap, 1rem);
  --_divider-color: var(--bslib-toolbar-divider-color, var(--bs-border-color-translucent, rgba(40,70,94,0.1)));
```

I'll let you decide if you think we should let users customize this in `toolbar_divider()`.

inst/components/scss/toolbar.scss
```diff
@@ -53,6 +59,24 @@
...
+    height: var(--_divider-height);
+    width: var(--_divider-gap);
+
+      // Use a pseudo-element for the actual line so custom width controls spacing
+      &::before {
+        content: "";
+        display: block;
+        width: var(--_divider-width);
+        height: 100%;
+        background-color: var(--bs-border-color-translucent);
```
```suggestion
        background-color: var(--_divider-color);
```

inst/components/scss/toolbar.scss
elnelson575: Works for me-- I think we should leave it as a CSS-only change to keep the args a bit slimmer. If someone wants to change it on a toolbar-by-toolbar basis in the code, they can pass it in like:
```
        toolbar(
          style = "--_divider-color: red;",
          toolbar_input_button(
            id = "btn1f",
            label = "Home",
            icon = icon("home")
          ),
          toolbar_divider(), # Default divider
```

Issue #1271, opened 2025-12-18
Release bslib 0.10.0


gadenbuie:
Prepare for release:

* [x] `git pull`
* [x] Check [current CRAN check results](https://cran.rstudio.org/web/checks/check_results_bslib.html)
* [x] [Advance deprecations](https://lifecycle.r-lib.org/articles/communicate.html#gradual-deprecation), if needed
* [ ] [Polish NEWS](https://style.tidyverse.org/news.html#news-release)
* [x] `urlchecker::url_check()`
* [ ] `devtools::build_readme()`
* [x] `devtools::check(remote = TRUE, manual = TRUE)`
* [ ] `devtools::check_win_devel()`
* [x] `revdepcheck::cloud_check()`
* [ ] Update `cran-comments.md`
* [ ] `git push`
* [ ] Draft blog post
* [ ] Update static imports: `staticimports::import()`

Submit to CRAN:

* [x] `usethis::use_version('minor')`
* [ ] `devtools::submit_cran()`
* [ ] Approve email

Wait for CRAN...

* [ ] Accepted :tada:
* [ ] Finish & publish blog post
* [ ] Add link to blog post in pkgdown news menu
* [ ] `usethis::use_github_release()`
* [ ] `usethis::use_dev_version(push = TRUE)`
* [ ] Share on social media
* [ ] Slack link to blog post, bluesky, and linkedin in #open-source-comms

PR #1272, opened 2025-12-18
v0.10.0

gadenbuie:
For #1271


gadenbuie:
revdepchecks revealed a new failure in OmopViewer, will be fixed by
* https://github.com/OHDSI/OmopViewer/pull/355

PR #1273, opened 2025-12-30
Add input bindings for toolbar buttons and select components

### LINKED ISSUE #1266: Toolbar: Add input bindings for updateActionButton and updateSelectInput
Current usage of the input bindings doesn't permit `updateActionButton` to function properly and does not work at all for `updateSelectInput`.

We should fix this so these work well in toolbars.

elnelson575:
Fixes #1266 

Objectives:

Create input for single select and action button, including update functions.

Todo: Create an issue to deal with validation as specified below
**Decision Log:**

-  Update tooltip will remain a separate function call, not a part of update_toolbar_*()

-  **Resolved. Question**: Right now, if the options are changed using `update_toolbar_input_select()`, there is not a selected choice set by default unless it is also specified using the selected argument. This is consistent with the standard `input_select()`s behavior. An alternative would be to automatically select the new choices' first value on update, clearing any selection already made.
- If current choice is not in choices, clear it and select the first in the new choice list
- Implemented a small mocking function for testing the errors with blank labels

**Additional test and failure cases:**

**Select Input:**
-  If a user attempts to update a label to "" they will get a fatal warning

**Input Action Button:**
-  If a user attempts to update a label to "" they will get a non-fatal warning

**Examples and recordings:**
<details>
<summary>Action Button Test App</summary>

https://github.com/user-attachments/assets/56969a09-f5c0-408c-9b80-e54b571abb57

```
library(shiny)
library(bslib)

ui <- page_fluid(
  h2("Toolbar Button Update Example"),
  p("This example demonstrates updating toolbar buttons dynamically."),

  card(
    card_header(
      "Media Player",
      toolbar(
        align = "right",
        toolbar_input_button(
          "play",
          label = "Play",
          icon = icon("play"),
          show_label = FALSE
        ),
        toolbar_input_button(
          "stop",
          label = "Stop",
          icon = icon("stop"),
          show_label = FALSE,
          disabled = TRUE
        ),
        toolbar_divider(),
        toolbar_input_button(
          "volume",
          label = "Volume: High",
          icon = icon("volume-high"),
          show_label = FALSE
        )
      )
    ),
    card_body(
      h4("Player Status"),
      verbatimTextOutput("status"),
      hr(),
      h4("Button Click Counts"),
      verbatimTextOutput("clicks"),
      actionButton("dummy", "Dummy")
    )
  ),

  card(
    card_header(
      "Another Example: Task Status",
      toolbar(
        align = "right",
        toolbar_input_button(
          "hide_label",
          label = "Boo!",
          icon = icon("exclamation-triangle"),
          show_label = FALSE
        )
      )
    ),
    card_body(
      toolbar(
        align = "left",
        toolbar_input_button(
          "task",
          label = "Do a Task",
          icon = icon("play-circle"),
          show_label = TRUE
        )
      ),
      verbatimTextOutput("task_status")
    )
  )
)

  # Track player state
  playing <- reactiveVal(FALSE)
  volume_level <- reactiveVal(3) # 1=mute, 2=low, 3=high
  label_shown <- reactiveVal(FALSE) # Track hide_label button state

  # Handle play button
  observeEvent(input$play, {
    will_be_playing <- !playing()
    update_toolbar_input_button(
      "play",
      label = if (will_be_playing) "Pause" else "Play",
      icon = icon(if (will_be_playing) "pause" else "play"),
      session = session
    )
    update_tooltip(
      "play_tooltip",
      if (will_be_playing) "Pause playback" else "Start playback",
      session = session
    )
    update_toolbar_input_button(
      "stop",
      disabled = !will_be_playing,
      session = session
    )
    playing(will_be_playing)
  })

  # Handle stop button
  observeEvent(input$stop, {
    playing(FALSE)
    update_toolbar_input_button(
      "play",
      label = "Play",
      icon = icon("play"),
      session = session
    )
    update_tooltip(
      "play_tooltip",
      "Start playback",
      session = session
    )
    update_toolbar_input_button(
      "stop",
      disabled = TRUE,
      session = session
    )
  })

  # Handle volume button - cycle through mute, low, high
  observeEvent(input$volume, {
    current <- volume_level()
    new_level <- if (current == 3) 1 else current + 1
    volume_level(new_level)

    vol_info <- switch(
      new_level,
      list(icon = "volume-xmark", label = "Volume: Mute"),
      list(icon = "volume-low", label = "Volume: Low"),
      list(icon = "volume-high", label = "Volume: High")
    )

    update_toolbar_input_button(
      "volume",
      label = vol_info$label,
      icon = icon(vol_info$icon)
    )
    update_tooltip("volume_tooltip", vol_info$label)
  })

  # Show/hide label button
  observeEvent(input$hide_label, {
    new_state <- !label_shown()
    update_toolbar_input_button(
      "hide_label",
      show_label = new_state,
    )
    label_shown(new_state)
  })

  # Handle task button
  observeEvent(input$task, {
    # Update to show task is complete
    update_toolbar_input_button(
      "task",
      label = "Task Complete!",
      icon = icon("check-circle"),
      session = session
    )

    # Reset to original after 2 seconds
    later::later(
      function() {
        update_toolbar_input_button(
          "task",
          label = "Do a Task",
          icon = icon("play-circle"),
          session = session
        )
      },
      delay = 2
    )
  })

  # Display status
  output$status <- renderPrint({
    cat("Player State:", if (playing()) "Playing" else "Stopped", "\n")
    vol_text <- switch(
      volume_level(),
      "Muted",
      "Low",
      "High"
    )
    cat("Volume Level:", vol_text, "\n")
  })

  output$clicks <- renderPrint({
    cat("Play button clicks:", input$play, "\n")
    cat("Stop button clicks:", input$stop, "\n")
    cat("Volume button clicks:", input$volume, "\n")
    cat("Regular button clicks:", input$dummy, "\n")
  })

  output$task_status <- renderPrint({
    cat("Task completed", input$task, "time(s)\n")
  })
}

shinyApp(ui, server)


```

</details>

<details>
<summary>Select Test App</summary>

https://github.com/user-attachments/assets/4f9e212c-ef4f-437a-af4f-816796344b5e

```
library(shiny)
library(bslib)

ui <- page_fluid(
  h2("Toolbar Input Select Test App"),

  card(
    card_header(
      "Toolbar with Select Input",
      toolbar(
        align = "right",
        toolbar_input_select(
          "filter",
          label = "Filter",
          choices = c("All", "Active", "Inactive"),
          selected = "All",
          icon = icon("filter")
        ),
        toolbar_input_button(
          "refresh",
          label = "Refresh",
          icon = icon("arrows-rotate")
        )
      )
    ),
    card_body(
      h4("Current Selection"),
      verbatimTextOutput("selection"),
      hr(),
      h4("Update Controls"),
      fluidRow(
        column(
          6,
          textInput("new_label", "New Label", value = "Filter"),
          checkboxInput("show_label_check", "Show Label", value = FALSE),
          actionButton(
            "update_label",
            "Update Label & Visibility",
            class = "btn-primary"
          )
        ),
        column(
          6,
          selectInput(
            "new_icon",
            "New Icon",
            choices = c("filter", "search", "list", "table", "chart-bar"),
            selected = "filter"
          ),
          actionButton("update_icon", "Update Icon", class = "btn-primary")
        )
      ),
      hr(),
      fluidRow(
        column(
          6,
          radioButtons(
            "choice_set",
            "Choice Set",
            choices = c(
              "Status" = "status",
              "Priority" = "priority",
              "Category" = "category"
            ),
            selected = "status"
          ),
          actionButton(
            "update_choices",
            "Update Choices",
            class = "btn-primary"
          )
        ),
        column(
          6,
          h5("Choice Sets:"),
          tags$ul(
            tags$li(tags$strong("Status:"), " All, Active, Inactive"),
            tags$li(tags$strong("Priority:"), " Low, Medium, High, Critical"),
            tags$li(
              tags$strong("Category:"),
              " Feature, Bug, Enhancement, Documentation"
            )
          )
        )
      ),
      hr(),
      h4("Test All Updates Together (All new values)"),
      actionButton(
        "update_all",
        "Update Everything",
        class = "btn-success btn-lg"
      )
    )
  )
)

  # Reactive value that updates whenever filter changes
  filter_info <- reactive({
    list(
      value = input$filter,
      timestamp = Sys.time()
    )
  })

  # Display current selection
  output$selection <- renderPrint({
    filter_info()
  })

  # Update label and show_label
  observeEvent(input$update_label, {
    update_toolbar_input_select(
      "filter",
      label = input$new_label,
      show_label = input$show_label_check
    )
    update_tooltip(
      "filter-tooltip",
      input$new_label
    )
    showNotification(
      sprintf(
        "Updated label to '%s' (visible: %s) and tooltip",
        input$new_label,
        input$show_label_check
      ),
      type = "message"
    )
  })

  # Update icon
  observeEvent(input$update_icon, {
    update_toolbar_input_select(
      "filter",
      icon = icon(input$new_icon)
    )
    showNotification(
      sprintf("Updated icon to '%s' and tooltip", input$new_icon),
      type = "message"
    )
  })

  # Update choices based on selected set
  observeEvent(input$update_choices, {
    new_choices <- switch(
      input$choice_set,
      "status" = c("All", "Active", "Inactive"),
      "priority" = c("Low", "Medium", "High", "All", "Critical"),
      "category" = c("Feature", "Bug", "All", "Enhancement", "Documentation")
    )

    update_toolbar_input_select(
      "filter",
      choices = new_choices,
      selected = "Low"
    )
    showNotification(
      sprintf("Updated choices to '%s' set", input$choice_set),
      type = "message"
    )
  })

  # Update everything at once
  observeEvent(input$update_all, {
    update_toolbar_input_select(
      "filter",
      label = "Chaos",
      show_label = TRUE,
      icon = icon("star"),
      choices = c("Option A", "Option B", "Option C", "Option D"),
      selected = "Option B"
    )
    update_tooltip(
      "filter-tooltip",
      "Chaos"
    )
    showNotification(
      "Updated label, visibility, icon, choices, selection, and tooltip!",
      type = "message",
      duration = 3
    )
  })

  # Demonstrate dynamic updates based on button clicks
  observeEvent(input$refresh, {
    current <- input$filter
    showNotification(
      sprintf("Refreshing with filter: %s", current),
      type = "message"
    )

    # Simulate a refresh action
    Sys.sleep(0.5)

    showNotification(
      "Refresh complete!",
      type = "message"
    )
  })
}

shinyApp(ui, server)


```

</details>

### DIFF
--- a/R/toolbar.R
+++ b/R/toolbar.R
@@ -355,7 +355,7 @@ toolbar_input_select <- function(
     id = id,
     class = "form-select form-select-sm",
     `data-shiny-no-bind-input` = NA,
-    HTML(processed$options)
+    processed$options
   )
 
   # Add optional icon before the select
@@ -597,11 +597,11 @@ process_choices_selected <- function(
 
   # Process choices into HTML options
   if (!is.null(choices_normalized)) {
-    options_html <- as.character(selectOptions(
+    options_html <- HTML(as.character(selectOptions(
       choices_normalized,
       selected_validated,
       inputId = inputId
-    ))
+    )))
   }
 
   # Process selected value


R/toolbar.R
```diff
@@ -245,18 +352,22 @@ toolbar_input_select <- function(
   choicesWithNames <- asNamespace("shiny")[["choicesWithNames"]]
   choices <- choicesWithNames(choices)
 
+  # Use a unique ID for the select element to avoid conflicts with standard
+  # select binding. The wrapper will have the user-specified id.
```
gadenbuie: Can you elaborate a bit about what conflicts arise? (Here in this thread first)

R/toolbar.R
```diff
@@ -153,12 +153,119 @@ toolbar_input_button <- function(
...
+#' @export
+update_toolbar_input_button <- function(
+  id,
+  label = NULL,
+  show_label = NULL,
+  icon = NULL,
+  disabled = NULL,
+  session = get_current_session()
+) {
+  # Validate that label has text for accessibility
```
gadenbuie: I think you could remove all of the comments in this function

R/toolbar.R
```diff
@@ -153,12 +153,119 @@ toolbar_input_button <- function(
...
+  label_text <- paste(unlist(find_characters(label)), collapse = " ")
+  # Verifies the label contains non-empty text
+  if (!nzchar(trimws(label_text))) {
+    warning(
```
gadenbuie: Just for consistency

```suggestion
    rlang::warn(
```

R/toolbar.R
```diff
@@ -153,12 +153,119 @@ toolbar_input_button <- function(
...
+    NULL
+  }
+
+  # Process icon - wrap it in the same structure as toolbar_input_button()
+  # The icon content will be updated within the existing .bslib-toolbar-icon span
+  icon_processed <- if (!is.null(icon)) {
+    processDeps(validateIcon(icon), session)
+  } else {
+    NULL
+  }
```
gadenbuie: A small re-ordering; I think this form scans a little better.

```suggestion
  icon <- validateIcon(icon)
  icon_processed <- if (!is.null(icon)) processDeps(icon, session)
  label_processed <- if (!is.null(label)) processDeps(label, session)
```

R/toolbar.R
```diff
@@ -153,12 +153,119 @@ toolbar_input_button <- function(
...
+    disabled = disabled
+  ))
+
+  session$sendInputMessage(id, message)
+}
+
+# Input handler for toolbar_input_button
+toolbar_input_button_input_handler <- function(value, shinysession, name) {
+  # Match shinyActionButtonValue class so it behaves
+  # like a standard action button for event handlers and input validation
```
```suggestion
toolbar_input_button_input_handler <- function(value, shinysession, name) {
  # Treat like a standard action button for event handlers and input validation
```

R/toolbar.R
```diff
@@ -291,13 +402,135 @@ toolbar_input_select <- function(
...
+#' @inheritParams toolbar_input_select
+#' @param selected The new selected value. If `NULL`, the selection is not changed.
+#' @param session A Shiny session object (the default should almost always be
+#'   used).
+#'
+#' @details
+#' This update function works similarly to [shiny::updateSelectInput()], but
+#' is specifically designed for [toolbar_input_select()]. It allows you to
+#' update the select's label, icon, choices, selected value(s), and label
+#' visibility from the server.
```
gadenbuie: This part could be part of the regular description (shows above-the-fold in pkgdown docs)

R/toolbar.R
```diff
@@ -291,13 +402,135 @@ toolbar_input_select <- function(
...
+#'
+#' Note that you cannot change the `tooltip` parameter after the select has
+#' been created, as it affects the structure and ARIA attributes.
```
gadenbuie: I think this could be a bit more helpful. IIRC, we gave the tooltip a stable ID so that people could use `update_tooltip()`. We should include that advice.

That also implies we could be a little more specific: you can't disable or enable the tooltip once you've created the button, but you *can* update the text.

R/toolbar.R
```diff
@@ -291,13 +402,135 @@ toolbar_input_select <- function(
...
+  choices = NULL,
+  selected = NULL,
+  icon = NULL,
+  session = get_current_session()
+) {
+  # Label can be null if there is no update, but if it is supplied it must be
+  # valid
+  if (
+    !is.null(label) &&
+      (!is.character(label) || length(label) != 1 || !nzchar(trimws(label)))
```
```suggestion
    !(is.null(label) || (rlang::is_string(label) && nzchar(trimws(label))))
```

or 

```
if (!is.null(label)) {
  if (!(rlang::is_string(label) && nzchar(trimws(label)))) {
    rlang::abort("...")
  }
}
```

Basically, I'm willing to do a lot of work to avoid conditions in `if` statements that span more than one line. I always find it much easier to read and follow.

R/toolbar.R
```diff
@@ -291,13 +402,135 @@ toolbar_input_select <- function(
...
+  } else {
+    NULL
+  }
+
+  # Process icon if supplied
+  icon_processed <- if (!is.null(icon)) {
+    processDeps(validateIcon(icon), session)
+  } else {
+    NULL
+  }
```
gadenbuie: Same as above

```suggestion
  icon <- validateIcon(icon)
  icon_processed <- if (!is.null(icon)) processDeps(icon, session)
  label_processed <- if (!is.null(label)) processDeps(label, session)
```

R/toolbar.R
```diff
@@ -291,13 +402,135 @@ toolbar_input_select <- function(
...
+  # Follow Shiny's pattern: choices and selected are handled separately
+  choices_processed <- if (!is.null(choices)) {
+    # Normalize choices using util function imported from Shiny
+    choicesWithNames <- asNamespace("shiny")[["choicesWithNames"]]
+    choices <- choicesWithNames(choices)
+
+    # Generate the options HTML (selected will be marked in the HTML if provided)
+    options_html <- selectOptions(choices, selected, inputId = id)
+
+    as.character(options_html)
```
gadenbuie: This looks a bit like code duplication. Is it possible to move this into a helper function that we could call in both places?

That would also help us simplify the code to something like this

```r
choices_processed <- if (!is.null(choices)) select_choices(choices, selected, id)
selected_processed <- if (!is.null(selected)) as.character(selected)
```

(`select_choices()` is just a placeholder, I didn't think about it much)

R/utils-shiny.R
```diff
@@ -42,6 +42,10 @@ processDeps <- function(...) {
   getFromNamespace("processDeps", "shiny")(...)
 }
 
+validateIcon <- function(...) {
+  getFromNamespace("validateIcon", "shiny")(...)
+}
```
gadenbuie, line 47: Can you take a quick peak at the history of `validateIcon()`? For a long time it was notoriously picky and would only let `tags$i()` elements through. I know we finally changed that, but it was changed relatively recently.

srcts/src/components/toolbarInputButton.ts
```diff
@@ -0,0 +1,95 @@
...
+  icon?: string | { html: string; deps: HtmlDep[] };
+  disabled?: boolean;
+};
+
+/**
+ * This is a Shiny input binding for `bslib::toolbar_input_button()`.
+ * It extends the standard action button behavior with support for updating
+ * the button's label, icon, and disabled state.
+ */
+class BslibToolbarInputButtonBinding extends InputBinding {
```
gadenbuie, line 21: I wonder if we could import the input button binding directly to avoid having to define some of the basic methods?

srcts/src/components/toolbarInputButton.ts
```diff
@@ -0,0 +1,95 @@
...
+  async receiveMessage(el: HTMLElement, message: ToolbarInputButtonMessage) {
+    // Update disabled state
+    if (hasDefinedProperty(message, "disabled")) {
+      (el as HTMLButtonElement).disabled = message.disabled;
+    }
+
+    // Update label
+    if (hasDefinedProperty(message, "label")) {
+      const labelEl = el.querySelector(".bslib-toolbar-label") as HTMLElement;
+      if (labelEl && message.label !== undefined) {
```
gadenbuie: We always include the label and icon containers, right? Just checking that `if (labelEl)` and similar are about satisfying the type checkers

srcts/src/components/toolbarInputSelect.ts
```diff
@@ -0,0 +1,105 @@
...
+ * It extends the standard select input behavior with support for updating
+ * the select's label, icon, and choices.
+ */
+class BslibToolbarInputSelectBinding extends InputBinding {
+  find(scope: HTMLElement) {
+    return $(scope).find(".bslib-toolbar-input-select");
+  }
+
+  getId(el: HTMLElement): string {
+    // The wrapper element has the input ID
```
gadenbuie: I think you can remove all the internal comments in this class as well

tests/testthat/test-toolbar.R
```diff
@@ -713,3 +715,81 @@ test_that("bslib::selectOptions() matches shiny::selectOptions() output", {
...
+
+test_that("update_toolbar_input_select() validates label parameter", {
+  # Empty string label should error (validation happens before session is used)
+  expect_error(
+    update_toolbar_input_select(
+      "test_id",
+      label = ""
+    ),
+    "`label` must be a non-empty string"
+  )
```
gadenbuie: This is somewhat of a more recent style adjustment, but I like using `expect_snapshot()` with `error = TRUE` for these messages. That way, the error message is captured in the snapshot, and the expectation fails if no error is thrown, even on CRAN.

```suggestion
  expect_snapshot(error = TRUE, {
    update_toolbar_input_select("test_id", label = "")
  })
```

On the other hand, this approach introduces the problem that if `label = ""` made it past the validation, we'd get a different error about missing an active Shiny session. That could be fixed by mocking `session`, but I'm not sure it's worth it. Up to you!

It would be nice to have the update function on one line in these tests though to make it easier to scan


```suggestion
  expect_error(
    update_toolbar_input_select("test_id", label = ""),
    "`label` must be a non-empty string"
  )
```

R/toolbar.R
```diff
@@ -153,12 +153,119 @@ toolbar_input_button <- function(
...
+  id,
+  label = NULL,
+  show_label = NULL,
+  icon = NULL,
+  disabled = NULL,
+  session = get_current_session()
+) {
+  # Validate that label has text for accessibility
+  label_text <- paste(unlist(find_characters(label)), collapse = " ")
+  # Verifies the label contains non-empty text
```
gadenbuie: Given how we now handle `label` in `toolbar_input_select()`, I think we might want to bump this up to an error as well. I.e. require that `label` is provided for accessibility reasons, even if not shown.

Although now I'm remembering that the difference is that in the select input we could also enforce `label` being a scalar string, where here it could be tags that _do_ meet accessibility requirements. Hmmm 🤔

R/toolbar.R
elnelson575: Looking back at my notes, it looks like this is because we're wrapping this select element (with the label and icon) and we want to update those as well as the select element within it. We want any update message to go through the toolbar-specific select bindings, not the select binding, which won't know what to do with some of the params we have.
This way, we only register the wrapped select with the toolbar select bindings and avoid having the select double-claimed, if that makes sense?

elnelson575: From discussion: Revert back to the code at https://github.com/rstudio/bslib/pull/1273/changes/bdda4da733400760b43ddfafe8c6f034f8fdfa53

should register correctly with data-shiny-no-bind-input added to the select tag, but verify

elnelson575: This worked! Updated version to be pushed shortly.

srcts/src/components/toolbarInputButton.ts
elnelson575: For posterity, we investigated this and went NOPE, because it gets complicated based on what is exported from shiny, when it's loaded, etc. The current approach follows the same path as the long running task button.

R/toolbar.R
elnelson575: Added the stable tooltip id and clarified the text here. 👍🏻

R/toolbar.R
elnelson575: The choices one, definitely! The selected one is unfortunately slightly different in both places because of needing HTML vs. tag output. We could wrap it if we really wanted to, but for now I chose to leave that part as is.

R/utils-shiny.R
elnelson575: Looks like now it will accept non-tag-like values, it just warns if isTagLike is false (as of 1.11 in June 2024). icon, span, things marked like attr() should all go through without a warning.

srcts/src/components/toolbarInputButton.ts
elnelson575: Ah yep, good catch. We can remove this for this and the select.

R/toolbar.R
elnelson575: Yeah, I think that had been why we warned instead of errored, since we didn't want to make a strict string requirement on people's labels in the buttons because they're more likely to actually use visible labels, vs. in the select, we're assuming the vast majority of people will use no visible label, which means the tooltip/accessible annotations are way more important.

R/toolbar.R
```diff
@@ -153,12 +153,103 @@ toolbar_input_button <- function(
   if (!is.null(tooltip)) {
     # Default placement is "bottom" for the toolbar case because otherwise the
     # tooltip ends up covering the neighboring buttons in the header/footer.
-    button <- tooltip(button, tooltip, placement = "bottom")
+    button <- tooltip(
+      button,
+      tooltip,
+      id = sprintf("%s_tooltip", id),
```
gadenbuie: We should document this in `?toolbar_input_button()` too

R/toolbar.R
```diff
@@ -153,12 +153,103 @@ toolbar_input_button <- function(
...
+#' @inheritParams toolbar_input_button
+#' @param session A Shiny session object (the default should almost always be
+#'   used).
+#'
+#' @details
+#' This update function works similarly to [shiny::updateActionButton()], but
+#' is specifically designed for [toolbar_input_button()]. It allows you to
+#' update the button's label, icon, and disabled state from the server.
+#'
+#' Note that you cannot change `show_label`, `tooltip`, or `border` parameters
```
gadenbuie: We should document updating `tooltip` here as well.

Is `show_label` included here by accident? It's in the function signature, so it seems like you are intending to support changing label visibility? We should either update the docs or drop `show_label` from the update function.

R/toolbar.R
```diff
@@ -235,28 +326,32 @@ toolbar_input_select <- function(
   # Restore input for bookmarking
   selected <- shiny::restoreInput(id = id, default = selected)
 
-  # Set selected to the first choice if no default or restored value
+  choices <- normalize_choices(choices)
+
```
gadenbuie: Doesn't `process_choices_selected()` call `normalize_choices()`?

```suggestion
```

R/toolbar.R
```diff
@@ -235,28 +326,32 @@ toolbar_input_select <- function(
...
   firstChoice <- asNamespace("shiny")[["firstChoice"]]
   if (is.null(selected)) {
     selected <- firstChoice(choices)
   }
 
-  # Normalize choices using util function imported from Shiny
-  choicesWithNames <- asNamespace("shiny")[["choicesWithNames"]]
-  choices <- choicesWithNames(choices)
+  processed <- process_choices_selected(choices, selected, id)
```
gadenbuie: I think this line encompasses `firstChoice` and `selected` as well, right?

R/toolbar.R
gadenbuie: Also, we'll need to make `show_label` consistent across all components. So if we're enabling it in the select input, we'll need to make it work for the button, too. (If it doesn't already. Or vice versa.)

R/toolbar.R
```diff
@@ -286,18 +381,208 @@ toolbar_input_select <- function(
...
+  id,
+  label = NULL,
+  show_label = NULL,
+  choices = NULL,
+  selected = NULL,
+  icon = NULL,
+  session = get_current_session()
+) {
+  # Label can be null if there is no update, but if it is supplied it must be
+  # valid
```
```suggestion
  # Label can be null if there is no update, but if supplied it must be valid
```

R/toolbar.R
```diff
@@ -286,18 +381,208 @@ toolbar_input_select <- function(
...
+    values <- unname(choices_normalized)
+  }
+
+  as.character(values)
+}
+
+# Helper function to process and validate choices and selected
+# Returns a list with:
+#   - data: list with 'options' (HTML) and 'value' (selected value), or NULL entries
+#   - error: error message if validation failed (NULL if no error)
```
gadenbuie: Let's flatten this one level and have `process_choices_selected()` return `$options`, `$value` and `$error`

R/toolbar.R
```diff
@@ -286,18 +381,208 @@ toolbar_input_select <- function(
...
+      }
+    }
+  }
+
+  # Process choices into HTML options
+  options_html <- NULL
+  if (!is.null(choices_normalized)) {
+    options_html <- as.character(selectOptions(
+      choices_normalized,
+      validated_selected,
```
gadenbuie: small nit but `selected_valid` or `selected_validated` would be a more consistent name

R/toolbar.R
```diff
@@ -286,18 +381,208 @@ toolbar_input_select <- function(
...
+          "`selected` value '%s' is not in `choices`.",
+          as.character(selected)
+        )
+        validated_selected <- NULL
+      }
+    }
+  }
+
+  # Process choices into HTML options
+  options_html <- NULL
```
gadenbuie: I'd rather you init `options_html` and `value` to `NULL` first, then have two top-level `if` blocks

```r
options_html <- NULL
value <- NULL

if (...) {
  # options_html <- ...
}

if (...) {
  # value <- ...
}
```

R/toolbar.R
elnelson575: Agh, yes, good catch. Show/hide label already works on the button, this was a typo

R/toolbar.R
elnelson575: Yep, removing, good catch

elnelson575: Ah, now I remember why this was there-- firstChoice is only called in the case of the initial input creation (where we want to select the first choice vs. not send an update when selected = NULL) and firstChoice needs processed choices

elnelson575: If we combine that into the processing function, we can remove it here.

R/toolbar.R
elnelson575: No, because we only want to use the first choice if `selected` isn't provided in the *initial* setup, but not in the update.
Moved this into the processed helper function, but it does make it a bit more complicated. I honestly could go either way on burying this in there and avoiding processing choices twice vs. having it out here and keeping the processing function simpler.

gadenbuie: oh I see. 🤔 I do think it'd be cleaner to move everything into `process_choices_selected()`. Maybe you could have an argument `use_first_choice = FALSE` that is set to `TRUE` in the UI function?

R/toolbar.R
```diff
@@ -235,28 +327,32 @@ toolbar_input_select <- function(
...
+  if (!is.null(processed$error)) {
+    rlang::abort(processed$error)
+  }
 
   select_tag <- tags$select(
     id = id,
     class = "form-select form-select-sm",
-    selectOptions(choices, selected, inputId = id)
+    `data-shiny-no-bind-input` = NA,
+    HTML(processed$options)
```
gadenbuie: `HTML()` is basically equivalent to a `character` type; does anything break if we have `process_choices_selected()` mark `options` as `HTML()`?

elnelson575: I don't think it will break anything--testing now

elnelson575: Looks like it works!

PR #1276, opened 2026-01-13
Feat/toolbar epic

elnelson575:
This PR includes:

- Toolbar buttons + input bindings
- Toolbar select + input bindings
- Toolbar divider
- Toolbar container

**Punch List:**

- [x] Doc cleanup with toolbar_input_button() & update_toolbar_input_button() -- same page, no duplicate params, examples cleaned up/condensed, description revised to accommodate 
  - [x] Update describedIn
  - [x] Update params so no duplicates
  - [x] Condense examples
  - [x] Revise description
  - [x] Run example
- [x] Doc cleanup with toolbar_select_input() & update_toolbar_select_input()
  - [x] Update describedIn
  - [x] Update params so no duplicates
  - [x] Condense examples
  - [x] Revise description
  - [x] Run example
- [x] Test kitchen sink examples after merge in toolbar_spacer() to make sure nothing went whack (alas, it did so I changed `width` to default to "auto". The behavior that was the issue is below.
  - [x] Add the label version to this, check to make sure everything is in there
- [x] Add examples to the main toolbar using section (section: cookbook) -- look at value box website help page
  - [x] Card header
  - [x] Card footer
  - [x] Card labels
- [x] Remove toolbar functions from seeAlso() add things like card(), card_header() etc. as applicable
- [x] Do a final read through when it's all built

<details>
<summary>Issue with `width=100%`</summary>

https://github.com/user-attachments/assets/06ac7d29-a8b7-4b34-aa89-5c136c4d4670

</details>


R/toolbar.R
```diff
@@ -0,0 +1,707 @@
...
+#' @param ... Additional attributes to pass to the button.
+#' @param disabled If `TRUE`, the button will not be clickable. Use
+#'   [update_toolbar_input_button()] to dynamically enable/disable the button.
+#' @param border Whether to show a border around the button.
+#'
+#' @return Returns a button suitable for use in a toolbar.
+#'
+#' @family Toolbar components
+#' @export
+
```
```suggestion
```

R/toolbar.R
```diff
@@ -0,0 +1,707 @@
...
+  button
+}
+
+#' Update toolbar button input
+#'
+#' @description
+#' Change the value or appearance of a toolbar button input on the client.
+#'
+#' @rdname toolbar_input_button
+#' @inheritParams toolbar_input_button
```
gadenbuie: If the goal is to include `update_toolbar_input_button()` in the same documentation page as `toolbar_input_button()`, then typically you put all of the documentation in `toolbar_input_button()` and use `@describeIn` to keep the two together.

Currently the documentation is well set-up to have update be documented on its own and I'm fine with that, which would just mean dropping `@rdname`


```suggestion
#' @inheritParams toolbar_input_button
```

elnelson575: I think I was making this update as you were writing this comment 😂  all done!

R/toolbar.R
```diff
@@ -0,0 +1,923 @@
...
+  if (isFALSE(tooltip)) {
+    tooltip <- NULL
+  }
+  if (!is.null(tooltip)) {
+    # Default placement is "bottom" for the toolbar case because otherwise the
+    # tooltip ends up covering the neighboring buttons in the header/footer.
+    button <- tooltip(
+      button,
+      tooltip,
+      id = sprintf("%s-tooltip", id),
```
gadenbuie: We should use `"%s_tooltip"` here because users will access this through `input${id}_tooltip`. The general rule is `_{suffix}` for user-facing namespace insertions that are **documented**, or `-{suffix}` for internal namespace additions that we don't document and don't intend for users to use/see.

gadenbuie:
@elnelson575 we also need to add a section to `pkgdown/_pkgdown.yml` for toolbar

R/toolbar.R
elnelson575: Ah this is helpful, I was wondering about that. Will make that change.

pkgdown/_pkgdown.yml
```diff
@@ -55,6 +55,8 @@ navbar:
           href: articles/cards/index.html
         - text: Sidebars
           href: articles/sidebars/index.html
+        - text: Toolbars
+          href: reference/toolbar.html
```
gadenbuie: Oh, instead of a link here, we need to add toolbars to the `reference` section (below)

elnelson575: Ah, okay-- added and pushed.

R/toolbar.R
```diff
@@ -0,0 +1,922 @@
...
+#' A toolbar which can contain buttons, inputs, and other UI elements in a small
+#' form suitable for inclusion in card headers, footers, and other small places.
+#'
+#' bslib includes a complete example of toolbars and the many ways they can be used:
+#'
+#' ```r
+#' shiny::runExample("toolbar", package = "bslib")
+#' ```
+#'
+#' @section Toolbars in labels:
```
gadenbuie: Move into `@section Cookbook`

R/toolbar.R
```diff
@@ -0,0 +1,922 @@
...
+#'   ),
+#'   value = 1,
+#'   min = 0,
+#'   max = 100
+#' )
+#' ```
+#'
+#' @section Cookbook:
+#'
+#' Card headers are a common place you might want to use toolbars. Toolbars
```
gadenbuie: There's also a `@subsection` directive that I think would be useful here to give each example its own header

R/toolbar.R
```diff
@@ -0,0 +1,922 @@
...
+#' allow you to clearly show that a selection of inputs pertain to that
+#' particular card. For example, this card uses multiple
+#' `toolbar_input_select()` for filtering and sorting, along with toolbar
+#' buttons.
+#' ```r
```
gadenbuie: This is valid but the backticks get claustrophobic when they're forced to be so close to paragraphs (let theme breathe! 😄) Also below too.

R/toolbar.R
```diff
@@ -0,0 +1,922 @@
...
+#'   max = 100
+#' )
+#' ```
+#'
+#' @section Cookbook:
+#'
+#' Card headers are a common place you might want to use toolbars. Toolbars
+#' allow you to clearly show that a selection of inputs pertain to that
+#' particular card. For example, this card uses multiple
+#' `toolbar_input_select()` for filtering and sorting, along with toolbar
```
gadenbuie: Because this docstring is in `?toolbar`, function references to other help pages should be links


```suggestion
#' [toolbar_input_select()] for filtering and sorting, along with toolbar
```

R/toolbar.R
gadenbuie: or combine this iteration with the example you added

R/toolbar.R
```diff
@@ -0,0 +1,922 @@
...
+#' )
+#' ```
+#' You may also want to use toolbars in other places, such as inside input
+#' labels. For example:
+#' ```r
+#' card(
+#'   card_header(
+#'     "Text Editor Example"
+#'   ),
+#'   card_body(
```
gadenbuie: Let's have three examples:

1. My minimal example from the section above
2. This example, minus the `card()` chrome, i.e. just `textAreaInput()` with the extra buttons.
3. Another example using the new `input_textarea_submit()`

R/toolbar.R
```diff
@@ -0,0 +1,922 @@
...
+  as_fragment(tag)
+}
+
+#' Add a toolbar input button
+#'
+#' @description
+#' A button designed to fit well in small places such as in a [toolbar()].
+#'
+#' @section Updating toolbar buttons:
+#'   You can dynamically update the appearance and enabled/disabled state
```
gadenbuie: This formatting is commonplace in R documentation, but it's inconsistent with the section docs you wrote for `toolbar`. If it works fine, I think this is easier to maintain:




```suggestion
#' @section Updating toolbar buttons:
#' 
#' You can dynamically update the appearance and enabled/disabled state
```

R/toolbar.R
```diff
@@ -0,0 +1,922 @@
...
+}
+
+#' Toolbar Input Select
+#'
+#' @description
+#' Create a select input control that can be used to choose a single item from
+#' a list of values, suitable for use within a [toolbar()].
+#'
+#' @section Updating toolbar select inputs:
+#'   You can update the appearance and choices of a toolbar select input. This
```
gadenbuie: Same comment about section content indentation

R/toolbar.R
elnelson575: Added the text area, but kept a card header/footer example because otherwise we don't have any of the minimal/standard implementation in the cookbook, which feels a little odd

R/toolbar.R
elnelson575: Agree, I'll undo that indent.

R/toolbar.R
gadenbuie: Oh, sorry. My comment was about the toolbar-in-label form. I was seeing these as three variations on toolbar-in-label

R/toolbar.R
elnelson575: Apparently that's created using `##` at compile time, so used that ✔️

R/toolbar.R
elnelson575: Ahhhh okay, that makes more sense then! I'll update to reflect that.

PR #1277, opened 2026-01-15
feat: Support toolbars in input labels

### LINKED ISSUE #1259: Toolbar: Add toolbar spacer

gadenbuie:
Adds support for putting toolbars in Shiny input labels

* Adds `toolbar_spacer()` and fixes #1259
* Adds a small bit of CSS to make toolbars work in `<label>` elements


R/toolbar.R
```diff
@@ -662,11 +684,12 @@ selectOptions <- function(
...
 }
 
-#' Toolbar: Add a divider to a toolbar
+#' Toolbar: Add a divider or spacer to a toolbar
 #'
 #' @description
 #' `toolbar_divider()` creates a visual divider line with customizable width
-#' and spacing between toolbar elements.
+#' and spacing between toolbar elements. `toolbar_spacer()` creates empty space
+#' that expands to push adjacent toolbar elements apart as much as possible.
```
elnelson575: Trying to think if there's a different way to say this-- this kind of makes me feel like putting a spacer between two buttons in a toolbar should create a gap, but it doesn't, because there's no min size and the toolbar shrinks to fit its elements vs. flexing to be the full width of the header.
Don't know if we want to clarify that in some way, or say that spacer is more designed for labels in this spot too.

gadenbuie: Thanks, I'm pretty okay with the language as it is, hopefully the "push...as much as possible" acknowledges that that might be "not all".

For [nav_spacer()](https://rstudio.github.io/bslib/reference/nav-items.html) we just say

> `nav_spacer()`: Adding spacing between nav items.

gadenbuie: or maybe we could say "with customizable **fixed** width and spacing" for `toolbr_divider()`

gadenbuie: > this kind of makes me feel like putting a spacer between two buttons in a toolbar should create a gap, but it doesn't, because there's no min size and the toolbar shrinks to fit its elements vs. flexing to be the full width of the header.

Oh wait, we should fix this

PR #1290, opened 2026-02-25
Fix: Fix icon spacing and snapshot formatting

elnelson575:
**Summary**
Updates toolbar test snapshots to match the minified HTML output format introduced in shiny 1.12.1.9000.

**Background**
[Shiny PR #4350 ](https://github.com/rstudio/shiny/pull/4350) added .noWS parameters to actionButton() and actionLink() to fix an underline rendering bug. This change removes whitespace from the HTML output, producing HTML without whitespace instead of pretty-printed multi-line format.

Since toolbar_input_button() calls shiny::actionButton() internally, it inherits this no-WS HTML behavior. While this doesn't make much of a difference in shiny::actionButton()'s appearance, toolbar buttons include the tooltip wrapper, which has a larger surrounding html wrapper, so the block of html that is compacted is larger.

Additionally, the noWS inheritance seems to have caused icons in icon-only buttons to shift because how it interacts with flex containers and empty spans. Additional CSS rules were added to prevent this from being a problem in the future by not explicitly setting margins and flex vs. relying on inheritence.

**Changes**
Updates tests/testthat/_snaps/toolbar.md with de-WS'd HTML format (no other changes to HTML observed)
Updates icon SCSS to realign icons in icon-only buttons after Shiny updates caused them to float off to the side.


gadenbuie (approved):
LGTM thank you!

PR #1291, opened 2026-03-07
Add ggsql as a bundled language for input_code_editor

cpsievert:
## Summary

- Adds `ggsql` as a first-class language in prism-code-editor, so `input_code_editor(language="ggsql")` works natively
- The grammar extends SQL with ggsql-specific tokens: clause keywords (VISUALISE, DRAW, SCALE, etc.), geom types, aesthetic names, theme names, project types, and an extended SQL function list
- Grammar derived from the [TextMate grammar](https://github.com/posit-dev/ggsql/blob/main/ggsql-vscode/syntaxes/ggsql.tmLanguage.json)

## Files changed

- **`inst/lib/prism-code-editor/prism/languages/ggsql.js`** — new Prism grammar file that imports `sql.js` and extends it
- **`tools/yarn_install.R`** — added `"ggsql"` to `code_editor_bundled_languages`
- **`R/versions.R`** — added `"ggsql"` to the generated allowlist

## Test plan

- [x] `devtools::load_all(); shinyApp(page_fluid(input_code_editor("test", language = "ggsql", value = "SELECT * FROM t VISUALISE x AS x DRAW point")), function(input, output, session) {})` — verify VISUALISE, DRAW, point get distinct highlighting

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tools/build_ggsql_grammar.R
```diff
@@ -0,0 +1,258 @@
...
+    "",
+    "// Copy SQL tokens",
+    'Object.keys(sql).forEach(function(k) { ggsql[k] = sql[k]; });',
+    "",
+    "// ggsql clause keywords",
+    sprintf('ggsql["ggsql-keyword"] = {'),
+    sprintf("  pattern: %s,", format_word_regex(clause_keywords, "i")),
+    '  alias: "keyword",',
+    "};",
+    "",
```
gadenbuie: This part of the translation seems pretty brittle, in particular the way that `ggsl-geom` are inlined as strings.

I think it's worth setting this up in a way that it's as autonomous as possible. It's very likely that the person running this script will not have the context to recognize that it's broken or know what to do about it, unless we fail loudly and obviously.

OTOH, maybe there are reasons why this is safer than it looks

tools/build_ggsql_grammar.R
```diff
@@ -0,0 +1,258 @@
...
+    "// Geom types",
+    sprintf('ggsql["ggsql-geom"] = {'),
+    sprintf("  pattern: %s,", format_word_regex(geoms)),
+    '  alias: "builtin",',
+    "};",
+    "",
+    "// Scale type modifiers",
+    sprintf('ggsql["ggsql-scale-type"] = {'),
+    sprintf("  pattern: %s,", format_word_regex(scale_types, "i")),
```
gadenbuie: Small thing, but these would be easier to read if we explicitly name the `flags` argument.


```suggestion
    sprintf("  pattern: %s,", format_word_regex(scale_types, flags = "i")),
```

tools/build_ggsql_grammar.R
```diff
@@ -0,0 +1,258 @@
...
+    '  alias: "builtin",',
+    "};",
+    "",
+    "// Aesthetic names",
+    sprintf('ggsql["ggsql-aesthetic"] = {'),
+    sprintf("  pattern: %s,", format_word_regex(aesthetics)),
+    '  alias: "attr-name",',
+    "};",
```
gadenbuie: I think this whole section would be easier to read if we did it in sections of raw strings. (Also, seems like there's an unnecessary `sprintf()` happening here?)


```suggestion
    sprintf(
      r"(// Aesthetic names
ggsql["ggsql-aesthetic"] = {
  pattern: %s,
  alias: "attr-name"
};)",
      format_word_regex(aesthetics)
    ),
```

That said I don't have strong feelings, this doesn't need to be the prettiest file.

gadenbuie (approved):
Looks good to me, just one comment about the translation script looking a little brittle

tools/build_ggsql_grammar.R
```diff
@@ -0,0 +1,301 @@
...
+#   Rscript tools/build_ggsql_grammar.R
+#
+# This fetches the ggsql TextMate grammar from posit-dev/ggsql on GitHub,
+# extracts token patterns, and generates a Prism-compatible grammar file
+# that extends the built-in SQL grammar.
+#
+# The output is written to:
+#   inst/lib/prism-code-editor/prism/languages/ggsql.js
+
+TMGRAMMAR_URL <- "https://raw.githubusercontent.com/posit-dev/ggsql/main/ggsql-vscode/syntaxes/ggsql.tmLanguage.json"
```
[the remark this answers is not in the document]
cpsievert: We're going to keep it pointing at `main` — the script's `extract_token_group()` validation will fail loudly if the upstream grammar structure changes in a way we can't handle, so we get the benefit of tracking upstream automatically while still catching breaking changes. Pinning to a SHA would mean manual bumps for every upstream update with no real safety gain.

PR #1299, opened 2026-03-16
fix: improve sidebar resize handle for fine and coarse pointers

### LINKED ISSUE #1294: Allow me to disable the sidebar resize
The latest version allows sidebars to be resized with the mouse. It's a good feature, but it's not always wanted. There needs to be a way to turn it off, ideally with a `resizeable=FALSE` parameter.

gadenbuie:
Fixes #1294, both literally and in spirit.

## Summary

Improves the sidebar resize handle to work better across pointer types and to avoid conflicts with the sidebar's native scrollbar.

**Allow disabling resizable sidebars:** Several users have requested the ability to disable sidebar resizing, and this PR makes it possible by adding a `resizable` argument to `sidebar()`. The default is `TRUE`, so app authors who don't want resizable sidebars can now use `resizable = FALSE`.

**Handle width by pointer type:**
- Fine pointers (mouse): 8px (down from 12px)
- Coarse pointers (touch): 26px via `@media (any-pointer: coarse)`

**Scrollbar conflict resolution:**

The resize handle is positioned just inside the sidebar's outer edge, overlapping the native scrollbar. Previously, clicking the scrollbar would start a resize instead. This is resolved with a midpoint-crossing activation pattern:

- The handle has `pointer-events: none` by default, so the scrollbar beneath remains fully clickable
- A narrow 5px `::before` "trip area" at the sidebar's outer edge detects the mouse crossing the boundary
- When crossed, JS adds a `.handle-active` class that shows the `ew-resize` cursor (inherited by `::before`) and expands the trip area to 24px (12px per side)
- An engagement-point dismissal mechanism deactivates the handle if the mouse reverses back past where it activated, creating perceived directional bias without asymmetric sizing
- Touch users bypass this entirely — the handle is always interactive with a wide hit area

https://github.com/user-attachments/assets/65095785-af5b-4df9-95f4-8facc6f8bea7

## Verification

```r
library(shiny)
library(bslib)

ui <- page_sidebar(
  sidebar = sidebar(
    title = "Sidebar",
    lorem::ipsum(10)
  ),
  "Main content"
)

shinyApp(ui, function(input, output, session) {})
```

1. Hover over the sidebar's right edge — the resize cursor should NOT appear until the mouse crosses the sidebar boundary
2. Click and drag the sidebar scrollbar — it should scroll, not resize
3. Cross the sidebar edge toward the main area — the resize indicator and cursor should appear
4. Click and drag to resize the sidebar
5. Test on a touch device or with Chrome DevTools device emulation — the handle should be wider and always interactive


NEWS.md
```diff
@@ -2,6 +2,10 @@
 
 ## Improvements and bug fixes
 
+* Improved the sidebar resize handle to avoid conflicts with the sidebar's native scrollbar and to better support touch devices. (#1299)
+
+* `sidebar()` gains a `resizable` argument (default `TRUE`) to control whether the sidebar can be resized by dragging its edge on desktop (wide screen sizes). (#1299)
```
cpsievert, line 7: Should go under a new features section?

cpsievert (approved):

NEWS.md
gadenbuie: It's incremental and minor; I'd call it an "improvement" more than a "new feature"

PR #1307, opened 2026-05-08
fix: fix bs5 chevron on selectInput(selectize=FALSE)

### LINKED ISSUE #1295: selectInput(selectize=FALSE) requires a `form-select` class
This might belong in the shiny repo.

When creating a selectInput() with `selectize=FALSE` in bootstrap 5, the resulting input does not have a chevron arrow on the side that makes it clear this is an openable box. See screenshot. This makes it appear as a text input.

The input needs to add `class="form-select"` to appear correctly in bs5.

Image

elnelson575:
## Problem

`shiny::selectInput(selectize = FALSE)` renders `<select class="form-control">`, but in Bootstrap 5, `form-control` sets `appearance: none` without providing a `background-image` chevron. 

The result is that currently for Bootstrap 5, our `selectInput(selectize = FALSE)` is missing its dropdown arrow/chevron.

This bug occurs when the user both:
- Uses Bootstrap 5 (the bslib default)
- Have an `shiny::selectInput(selectize = FALSE)`

Fixes #1295

## Root cause

In BS4 and earlier, `appearance: none` on `<select>` was scoped only to date/time inputs. 
In BS5, it was applied unconditionally to all `.form-control` elements (to fix date inputs in Safari), which also strips the native browser chevron from selects. 
Bootstrap 5 introduced `.form-select` as the correct class for styled selects, but Shiny still uses the older `.form-control` for the default Bootstrap 3-based implementation.

## Fix

Added a rule to the bs3compat layer (`inst/bs3compat/_shiny_input.scss`, BS5 only) that applies the chevron `background-image` and supporting properties directly to `select.form-control`:

- Sets `--bs-form-select-bg-img` CSS custom property (the SVG chevron)
- Applies `background-image`, `background-repeat`, `background-position`, `background-size`
- Adds `padding-right` to make room for the chevron
- Suppresses the chevron for `[multiple]` and `[size]` selects so that the input behaves correctly for:
    - `multiple=True` — the select renders as a scrolling list box with several options visible at once, not as a collapsed dropdown.
  - `size` > 1 — any explicit size attribute (other than size="1") makes the browser render a multi-row list box instead of a dropdown
- Applies the dark-mode chevron variant when dark mode is enabled

The approach uses direct property declarations rather than `@extend .form-select` to avoid CSS bloat (extend would add `select.form-control` to every `.form-select` selector in the compiled output).

Also regenerated `inst/css-precompiled/5/bootstrap.min.css` — bslib serves this pre-built file for the default `bs_theme(version = 5)` without running Sass, so the fix must be included there too.

## Scope

The `select.form-control` selector only hits non-selectized inputs. When selectize takes over a select, it removes `form-control` and adds `selectized`, so this rule does not affect selectize widgets.

## Demo:

https://github.com/user-attachments/assets/59efc760-d045-4d6c-8fec-9863261d732f

<details>
<summary>App Code:</summary>
```

library(shiny)
library(bslib)

choices <- c("Option A", "Option B", "Option C")
bootswatch_choices <- c("default", bootswatch_themes(5))

ui <- page_fixed(
  title = "selectInput(selectize = FALSE) - Issue #1295",
  theme = bs_theme(version = 5),
  card(
    class = "mt-4",
    card_header(
      h2("selectInput(selectize = FALSE)", class = "mb-1"),
      p(
        class = "mb-0",
        "Native ",
        tags$code("<select>"),
        " elements rendered by Shiny with ",
        tags$code("class=\"form-control\""),
        " should display a chevron arrow in Bootstrap 5."
      )
    ),
    tags$style(HTML(
      "
      .d-flex > .shiny-input-container { margin-bottom: 0 !important; }
    "
    )),
    div(
      class = "d-flex align-items-end gap-3",
      style = "min-height: 80px;",
      selectInput(
        "theme",
        "Theme",
        choices = bootswatch_choices,
        width = "150px"
      ),
      div(class = "mb-2", input_dark_mode()),
    ),
    br(),
    br(),
    br()
  ),
  layout_columns(
    class = "mt-3",
    col_widths = c(4, 8),
    card(
      card_header("selectize = FALSE (native select)"),
      p(class = "text-success small mb-2", "Fixed by this PR"),
      selectInput("sel1", "Single select", choices, selectize = FALSE),
      p(class = "small bold mb-2", "Unaffected by this PR"),
      selectInput(
        "sel2",
        "Multiple select",
        choices,
        multiple = TRUE,
        selectize = FALSE
      ),
      selectInput("sel3", "Size = 3", choices, size = 3, selectize = FALSE)
    ),
    div(
      layout_column_wrap(
        width = 1 / 2,
        card(
          card_header("selectize = TRUE (default)"),
          p(class = "small bold mb-2", "Unaffected by this PR"),
          selectInput("sel4", "Single select", choices),
          selectInput("sel5", "Multiple select", choices, multiple = TRUE)
        ),
        card(
          card_header("textInput (for comparison)"),
          p(class = "small bold mb-2", "Unaffected by this PR"),
          textInput("txt", "Text input", placeholder = "Text...")
        )
      )
    )
  ),
  card(
    class = "mt-3",
    card_header(
      class = "d-flex align-items-center justify-content-between",
      "toolbar_input_select() - unaffected by this PR",
      toolbar(
        toolbar_input_select(
          "sel6",
          "Single select",
          choices,
          icon = icon("filter")
        )
      )
    ),
    div(class = "p-2")
  )
)

  observeEvent(input$theme, {
    theme <- if (input$theme == "default") {
      bs_theme(version = 5)
    } else {
      bs_theme(version = 5, bootswatch = input$theme)
    }
    session$setCurrentTheme(theme)
  })
}

shinyApp(ui, server)

```

</details>


inst/bs3compat/_shiny_input.scss
```diff
@@ -62,4 +62,30 @@ $form-group-margin-bottom: 1rem !default;
       margin-top: calc(#{-($line-height-base - $form-check-input-width) * .5} - #{$input-btn-border-width});
     }
   }
+
+  // selectInput(selectize=FALSE) uses form-control, but BS5's form-control sets
+  // appearance:none without a background-image chevron (unlike form-select).
+  // https://github.com/rstudio/bslib/issues/1295
```
gadenbuie: I found this comment hard to parse. I think what you're trying to say here is something like


```suggestion
  // selectInput(selectize=FALSE) uses .form-control, but in BS5 the select
  // chevron was moved to the new .form-select class.
```

I also think it'd be better to link to the source of these styles, which were presumably copied or migrated from _somewhere_, rather than to the issue, which doesn't really contain any new information beyond what's already in the comment.

elnelson575: Sorry, yeah this one is confusing. I tried to make it a little clearer in the update and also linked to the version of BS5 we're pulling the styles from. 👍🏻

elnelson575: (Also updated the PR description to be clearer because I realized a bit of context got lost in the details cut somehow)

gadenbuie (approved):
LGTM, thanks!

PR #1308, opened 2026-05-08
fix: elevate CSS specificity for BS5 radio/checkbox group label margin

elnelson575:
## Summary

- Bumps CSS specificity on bslib's BS5 radio/checkbox group label-spacing rule by doubling the class (`.shiny-input-radiogroup.shiny-input-radiogroup`)
- This makes bslib's rule win the specificity tie against Shiny's inline-specific override (`.shiny-input-radiogroup.shiny-input-container-inline`), so **inline** radio/checkbox groups now use the same calculated label-to-options spacing as their **stacked** counterparts.
- Note, for this to transfer to `py-shiny`, this needs to be merged into bslib, pyshiny needs the change in sass load order (so that this rule wins the specificity cascade), and then needs to revendor assets.

## What changes visually

Stacked (vertical / many-choice) groups are unchanged. Inline groups gain ~2px of breathing room between the group label and the first option so all three variants are consistent.

Computed `margin-top` on `.shiny-options-group`:

| | main | this PR |
|---|---|---|
| Vertical | −3.4px | −3.4px |
| Inline | **−1px** | **−3.4px** |
| Many | −3.4px | −3.4px |

### Before (main)

compare_main

### After (this PR)

compare_new_fix

## Test plan

- [x] Verify vertical `radioButtons()` label spacing is unchanged
- [x] Verify inline `radioButtons()` label spacing now matches vertical
- [x] Verify vertical `checkboxGroupInput()` label spacing is unchanged
- [x] Verify inline `checkboxGroupInput()` label spacing now matches vertical


inst/bs3compat/_shiny_input.scss
```diff
@@ -56,8 +56,8 @@ $form-group-margin-bottom: 1rem !default;
 // labels, see https://github.com/rstudio/shiny/blob/6fc06281/inst/www/shared/shiny_scss/shiny.scss#L344-L356
 // TODO: Consider removing if we add `input_checkbox()` and `input_radio()`
 @if $bootstrap-version == 5 {
-  .shiny-input-checkboxgroup,
-  .shiny-input-radiogroup {
+  .shiny-input-container.shiny-input-checkboxgroup,
+  .shiny-input-container.shiny-input-radiogroup {
```
gadenbuie: A trick that I like to use for increasing specificity without changing the rule is to duplicate the class name. It's unusual, but `.foo.foo` doubles the specificity of `.foo`. It also serves as a signal that the extra class is included only for specificity reasons, whereas adding an additional class could fail if the classes change and looks like it's included for semantic reasons.

elnelson575: Oooh tricky! Will update that!

cpsievert (approved):

PR #1310, opened 2026-05-11
chore: regenerate docs with roxygen2 8.0.0

elnelson575:
Pipeline Roxygen2 has updated to 8.0.0. This updates the current documentation to that format.

- [x] Checked to make sure site still renders properly

Doing this in a separate PR so as not to clutter the diffs of the current ones.

Note: Failure in routine is due to an issue with stringr (used to be installed as a transitory dependency, not anymore) and will be addressed in a separate PR.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -75,7 +75,6 @@ Config/testthat/start-first: zzzz-bs-sass, fonts, zzz-precompile, theme-*,
     rmd-*
 Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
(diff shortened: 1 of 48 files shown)


DESCRIPTION
gadenbuie, line 1: Can you run `usethis::use_tidy_description()` in the repo just to make sure this file is stable?

gadenbuie (approved):
Thanks!

DESCRIPTION
elnelson575: Yep, just did! 👍🏻

Issue #1326, opened 2026-06-29
Fails to render dynamic contents of popover if it does not have a title.


averissimo:
## Describe the problem

Fails to render contents of popover if it does not have a title.

This only happens since `shiny@1.4.0`.

```r
library(shiny)
ui <- bslib::page_fixed(
  theme = bslib::bs_theme("5"),
  bslib::card(class = "mt-5",
    bslib::card_header(
      bslib::tooltip(
        trigger = bslib::popover(
          trigger = shiny::icon("maximize"),
          id = "expbut",
          uiOutput("card_title"),
          uiOutput("card_yada")
        ),
        options = list(trigger = "hover"),
        class = "resize-button",
        "A tooltip"
      )
    ),
    "The card body..."
  )
)

server <- function(input, output) {
  default_slider_width <- reactiveVal(500)
  output$card_title <- renderUI({
    req(default_slider_width())
    shiny::tags$span("This is a new content")
  })
}

shiny::shinyApp(ui, server)
```

### Session Info

<details>
<pre><code>
─ Session info ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 setting  value
 version  R version 4.5.2 (2025-10-31)
 os       Ubuntu 26.04 LTS
 system   x86_64, linux-gnu
 ui       Positron
 language (EN)
 collate  en_US.UTF-8
 ctype    en_US.UTF-8
 tz       Europe/Lisbon
 date     2026-06-29
 pandoc   3.7.0.2 @ /usr/bin/pandoc
 quarto   1.6.42 @ /opt/quarto/bin/quarto

─ Packages ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 ! package     * version     date (UTC) lib source
   bslib         0.11.0.9000 2026-06-29 [1] Github (rstudio/bslib@c87abc5)
 P cachem        1.1.0       2024-05-16 [?] CRAN (R 4.5.0)
   cli           3.6.6       2026-04-09 [1] CRAN (R 4.5.2)
   devtools      2.5.2       2026-04-30 [1] CRAN (R 4.5.2)
 P digest        0.6.39      2025-11-19 [?] CRAN (R 4.5.2)
   ellipsis      0.3.3       2026-04-04 [1] CRAN (R 4.5.2)
 P fastmap       1.2.0       2024-05-15 [?] CRAN (R 4.5.0)
[...]

PR #1327, opened 2026-06-29
fix(popover): render dynamic outputs in title-less popover/tooltip

gadenbuie:
For #1326

Dynamically-rendered outputs (e.g. `uiOutput()`) inside a `popover()` or `tooltip()` **with no title** failed to render — the output stayed stuck "recalculating" and never appeared. (With a title, the same output renders fine.)

## Cause

This is an upstream Shiny regression introduced in shiny 1.14.0 (rstudio/shiny#4388). Shiny's `IntersectionObserver`-based visibility detection ends up targeting bslib's boxless `display:contents` popover/tooltip wrapper, so it never fires when the popover opens. The output's `ResizeObserver` is then the only trigger left, and it only fires if the output's box size changes — which, in a title-less popover, it doesn't (the empty output measures 0×0 both before and after opening).

Full root-cause analysis and the proposed Shiny-side fix are tracked in rstudio/shiny#4406.

## Fix (temporary workaround)

Add a tiny CSS rule giving shiny outputs inside popovers/tooltips a `min-width: 1px`, which guarantees a non-zero box so the `ResizeObserver` fires when the popover opens. Once the output has content it becomes `display:contents`, so the rule is a no-op — no visual impact.

- New partial `inst/components/scss/tooltip-popover.scss`
- Regenerated `inst/components/dist/components.css` via `tools/compile_component_sass.R`
- NEWS.md entry

This is a workaround pending the upstream Shiny fix (rstudio/shiny#4406); once that ships, this rule can be removed.


gadenbuie:
@copilot resolve the merge conflicts in this pull request

cpsievert (approved):

PR #1338, opened 2026-07-17
Split docs of `value_box()` and `showcase_*()` functions

### LINKED ISSUE #1336: Consider splitting the docs of `value_box()` and `showcase_*()` functions?
Hello, this is a request to clarify the [documentation of "Value box"](https://rstudio.github.io/bslib/reference/value_box.html). This page contains the definition of five functions that can be split in two categories:

- value boxes functions: `value_box()` and `value_box_theme()`
- showcase functions: `showcase_left_center()`, `showcase_top_right()`, `showcase_bottom()`

These two groups share relatively few arguments but in cumulative there are 16 arguments documented. In several occasions I thought one of the arguments could be used in one the groups of functions mentioned above when it actually didn't. This is not a big problem, it is just annoying to go back and forth between the function definition and the argument list to know which argument is related to which function.

Besides, the showcase functions are not present at all in examples. Would you consider moving the showcase functions to their own page and link to this page from the value box page?

etiennebacher:
This moves the three `showcase_*()` functions to their own page to improve the readability of `value_box()` docs. Technically, this change is visible to users but I'm not convinced it warrants a bullet point in `NEWS.md`. Let me know if you'd like me to add one.

Ping @gadenbuie 

Fixes #1336


gadenbuie:
Thanks @etiennebacher! Looks good, but can you fix the R CMD check issues?

etiennebacher:
@gadenbuie those should be fixed now but apparently CI workflows don't run on every commit?

gadenbuie (approved):
Thanks again!

PR #1342, opened 2026-08-10
feat(navs): Use the navset id as the tabset id

schloerke:
## Summary

When a navset has a user-supplied `id=`, use it as the `data-tabsetid` (and in the derived `tab-<tabsetid>-<index>` pane ids) instead of minting a random integer.

**Before:** `data-tabsetid="4785"` / `id="tab-4785-1"`
**After:** `data-tabsetid="my_tabs"` / `id="tab-my_tabs-1"`

This makes the rendered markup reproducible across renders and straightforward to target from custom CSS and JavaScript.

This is the R counterpart of posit-dev/py-shiny#2410, which makes the same change in Shiny for Python. `buildTabset()` is the single place tabset ids are minted — `shiny::tabsetPanel()`, `navbarPage()`, `navlistPanel()` and `tabPanel()` all delegate here, so no change is needed in rstudio/shiny.

## The selector-safety caveat

py-shiny gets this for free because `validate_id()` restricts input ids to `^\.?\w+$`. **R has no equivalent** — `shiny::NS()` just pastes, and there's no validation on the leaf id. So `tabsetPanel(id = "my.tabs")` is legal R today, and naively reusing it would emit `href="#tab-my.tabs-1"`, which a browser parses as `#tab-my` + `.tabs-1` → tab switching silently breaks.

So the id is only reused when it matches `^[A-Za-z0-9_-]+$` (`-` included because that's `ns.sep`). Anything else falls back to the random integer, i.e. to exactly today's behavior. Anonymous navsets and `nav_menu()` dropdowns keep the random id as before.

## Changes

- **`R/navs-legacy.R`** New internal `tabset_id()` helper; `buildTabset()` calls it instead of `p_randomInt()` directly.
- **`tests/testthat/test-navs-legacy.R`** Tests for `tabset_id()` and for the rendered markup: id reuse, stability across renders, module-style `mod-tabs` ids, the anonymous fallback, `nav_menu()` keeping its own id, and the unsafe-id fallback.
- **`NEWS.md`** Entry under a new Improvements section.

## Verification

- Full suite: **1152 passing**, no snapshot changes (no existing bslib snapshot contains `data-tabsetid`). The 2 failures in `test-rmd-skeletons.R` are a local environment gap — `thematic` isn't installed — and are unrelated.
- Confirmed the new tests are load-bearing: reverting `buildTabset()` to `p_randomInt()` fails 6 of them.


R/navs-legacy.R
```diff
@@ -625,6 +625,29 @@ navbarMenuTextFilter <- function(text) {
...
+# as "#tab-my" + ".tabs-0". Ids that don't qualify fall back to the random
+# integer, i.e. to the behavior from before this was introduced. `-` is allowed
+# because that is what `shiny::NS()` uses to join module namespaces.
+tabset_id <- function(id = NULL) {
+  is_selector_safe <- is.character(id) &&
+    length(id) == 1 &&
+    !is.na(id) &&
+    grepl("^[A-Za-z0-9_-]+$", id)
+
+  if (is_selector_safe) id else as.character(p_randomInt(1000, 10000))
```
schloerke:
```suggestion
  if (is_selector_safe) {
    id
  } else {
    as.character(p_randomInt(1000, 10000))
  }
```

R/navs-legacy.R
```diff
@@ -625,6 +625,29 @@ navbarMenuTextFilter <- function(text) {
...
+# straightforward to target from custom CSS/JS.
+#
+# Shiny does not restrict the characters in an input id, so an id is only safe
+# here if it survives being embedded in a CSS selector -- `tabsetPanel(id =
+# "my.tabs")` would otherwise emit `href="#tab-my.tabs-0"`, which a browser reads
+# as "#tab-my" + ".tabs-0". Ids that don't qualify fall back to the random
+# integer, i.e. to the behavior from before this was introduced. `-` is allowed
+# because that is what `shiny::NS()` uses to join module namespaces.
+tabset_id <- function(id = NULL) {
+  is_selector_safe <- is.character(id) &&
```
schloerke: If not a safe selector and an id is supplied, we should produce a message to the author

R/navs-legacy.R
```diff
@@ -625,6 +625,29 @@ navbarMenuTextFilter <- function(text) {
...
+# When the tabset has an input `id`, reuse it: input ids are already unique, so
+# the ids stay unique, and the rendered markup becomes stable across renders and
+# straightforward to target from custom CSS/JS.
+#
+# Shiny does not restrict the characters in an input id, so an id is only safe
+# here if it survives being embedded in a CSS selector -- `tabsetPanel(id =
+# "my.tabs")` would otherwise emit `href="#tab-my.tabs-0"`, which a browser reads
+# as "#tab-my" + ".tabs-0". Ids that don't qualify fall back to the random
+# integer, i.e. to the behavior from before this was introduced. `-` is allowed
+# because that is what `shiny::NS()` uses to join module namespaces.
```
schloerke:
```suggestion
```

schloerke: Removed in e9bd4b8c. The paragraph is largely redundant now that the fallback announces itself at runtime (see the sibling thread) — what's left is just why the id gets reused at all.

R/navs-legacy.R
schloerke: Applied in e9bd4b8c. The block form is also what made room for the warning in the `else` branch.

R/navs-legacy.R
schloerke: Done in e9bd4b8c — `rlang::warn()`, classed `bslib_tabset_id_unsafe` so it can be caught or selectively suppressed:

```
Warning: Not using the tabset `id` ("my.tabs") as its `data-tabsetid`.
i The `id` is embedded in a CSS selector, so it may only contain letters, numbers, `_`, and `-`.
i A random tabset id was used instead.
```

It fires only when an `id` was actually supplied — `id = NULL` (an anonymous navset or a `nav_menu()` dropdown) stays silent, since there's nothing the author could have done differently.

I went with `warn()` rather than `inform()` because the author asked for a specific `id` and silently didn't get it. Happy to downgrade to a message if you'd rather it not surface in `options(warn = 2)` setups.

Verified nothing in shiny trips it: `test-tabPanel.R` is the only file that builds tabsets and none of them pass an `id`, so shiny's suite is still 3130 passing / 0 failing against this branch.

schloerke: Reverted in 823c87ee — per Carson, dropping the selector-safety check entirely, and the warning goes with it.

That collapses the helper to a single expression, so it's inlined back into `buildTabset()`:

```r
tabsetId <- id %||% p_randomInt(1000, 10000)
```

The R source diff is now just that one line plus a short comment. The helper's unit tests and the warning tests are gone; the rendered-markup tests remain, plus one covering a module-namespaced id (`NS("mod")("tabs")` -> `data-tabsetid="mod-tabs"`).

Suites are green: bslib 1141 passing (the 2 `test-rmd-skeletons.R` failures are a local missing-`thematic` gap, not from this branch), and shiny 3130 passing against this branch with `_snaps/tabPanel.md` unchanged.

For the record, so it's findable later: an id that isn't selector-safe (e.g. `tabsetPanel(id = "my.tabs")`) now yields `href="#tab-my.tabs-1"`, which a browser reads as `#tab-my` + `.tabs-1`. Shiny doesn't restrict input id characters, so that's reachable — just accepted rather than guarded.

schloerke:
@cpsievert 

Main change: https://github.com/rstudio/bslib/pull/1342/changes#diff-4a94d77a54a8563907e126ed292e831fa5a7b40e01ddd56bdd8847e5624bbdc6

The rest is mainly automated file updates

cpsievert (approved):

PR #1353, opened 2026-09-08
tools, utils are Imports, not Suggests

MichaelChirico:
https://github.com/rstudio/bslib/blob/6935d9819fcb37e0b42ffa54f4e1cab0418ec2ce/NAMESPACE#L218-L226

See https://bugs.r-project.org/show_bug.cgi?id=19160


cpsievert:
Thanks!

